PR #18, opened 2026-07-21
build(deps): kotlin 2.4.10 + shadow 9.6.0 with regenerated verification metadata

marcospaulo:
The two Gradle bumps Dependabot can't land alone (each fails strict dependency verification). Batched because they share one `verification-metadata.xml` regeneration — the tricky part is the `:fir-checks` compiler-plugin **detached configuration**, whose kotlin-stdlib POM a `help`-task regen silently omits (that was the earlier CI failure). Regenerated with `--write-verification-metadata sha256 clean check :app:shadowJar`. Full gate + clean release rebuild green.

Closes #10, closes #11.


marcospaulo:
Splitting: shadow 9.6.0 lands as its own PR; **kotlin 2.4.10 is blocked by GitHub's CodeQL Kotlin extractor** (`KotlinVersionTooRecentError: 2.4.10 is too recent`), a required security check we won't bypass. Kotlin deferred until CodeQL adds 2.4.x support.

PR #22, opened 2026-07-22
fix(openai-chat): add emit_reasoning provider quirk for strict vendors

### LINKED ISSUE #21: openai-chat dialect always sends reasoning/reasoning_effort — strict OpenAI-compatible vendors (Fireworks) 400
## Summary

The `openai-chat` dialect unconditionally adds `reasoning` and `reasoning_effort` to every outbound Chat Completions request. OpenRouter tolerates the extension, but strict OpenAI-compatible vendors reject unknown fields. Against **Fireworks AI** every turn fails:

```
API Error: 400 Extra inputs are not permitted, field: 'reasoning'
```

There is currently no config-only way to disable it, so `openai-chat` api-key providers that validate strictly are unusable.

## Reproduce

1. Add an `openai-chat` provider for Fireworks:
   ```toml
   [providers.fireworks]
   dialect  = "openai-chat"
   base_url = "https://api.fireworks.ai/inference/v1"
   auth     = { kind = "api-key", env = "FIREWORKS_API_KEY" }
   [[providers.fireworks.models]]
   id = "accounts/fireworks/models/deepseek-v4-pro"
   label = "DeepSeek V4 Pro"
   context_window = 1048576
   ```
2. Add a head for it, `splice install --all`, export the key, restart.
3. Any turn → `400 Extra inputs are not permitted, field: 'reasoning'`.

## Root cause (commit dffe1bd)

- `dialect-openai-chat/.../ChatRequestBuilder.kt` (~L108–113) emits `reasoning_effort` and `reasoning`, gated only by `ChatQuirks.emitReasoningEffort`, whose default is `true` (~L38).
- `app/.../Daemon.kt` `chatProvider` (~L317–321) constructs `ChatQuirks` **without** reading `providerCfg.quirks`, so the flag can never be turned off from TOML.
- `QuirksConfig` (`core/.../Topology.kt` ~L101–109) has no reasoning key, and it is only overlaid onto the Responses dialect anyway.
- `show_reasoning = "off"` does **not** help: the chat builder ignores it for this field (only `ResponsesRequestBuilder` honors `isOff`, ~L317–318).

## Impact

No pasteable TOML makes `openai-chat` work with any strict OpenAI-compatible vendor (Fireworks, and likely others). The reference `openrouter` example works only because OpenRouter accepts the extension.

## Suggested fix

Expose `emitReasoningEffort` as a provider quirk (e.g. `emit_reasoning`, default `true` so existing configs are unchanged) and wire it into the chat `ChatQuirks` in `Daemon.kt`. Suppressing the request field is lossless: the mirror still reads `reasoning_content` from the response (`ChatStreamTranslator`), so DeepSeek/GLM reasoning still surfaces. PR to follow.

## Environment

splice 0.1.1 (commit dffe1bd), macOS/arm64, Java 22 runtime, Fireworks provider (`openai-chat`, api-key).

calube:
Fixes #21.

## What

Adds an `emit_reasoning` provider quirk (default `true`) that controls whether the `openai-chat` dialect sends the `reasoning` / `reasoning_effort` request fields, and wires it into the chat `ChatQuirks` construction.

## Why

Strict OpenAI-compatible vendors reject unknown request fields. Against Fireworks AI every turn currently fails with `400 Extra inputs are not permitted, field: 'reasoning'` (#21). There was no config-only way to disable the field — `QuirksConfig` had no reasoning key, and it is only overlaid onto the Responses dialect anyway.

## How

- `QuirksConfig` gains `@SerialName("emit_reasoning") emitReasoning: Boolean = true`.
- `Daemon.chatProvider` passes `emitReasoningEffort = providerCfg.quirks.emitReasoning` into both `ChatQuirks` branches (api-key and grok-oauth).

Default `true` preserves current behavior for OpenRouter/xAI. Setting `quirks = { emit_reasoning = false }` fixes Fireworks.

**Lossless:** the reasoning mirror still reads `reasoning_content` from the response (`ChatStreamTranslator`), so DeepSeek/GLM reasoning still surfaces — only the outbound request field is suppressed.

## Tests

- `EmitReasoningQuirkTest` (new): asserts `emit_reasoning = false` deserializes off the provider `quirks` table, and that it defaults to `true` when omitted.
- The existing `ChatRequestBuilderTest` already covers the builder omitting the fields when `emitReasoningEffort = false`.

## Verification

Built from this branch and ran end-to-end against Fireworks `deepseek-v4-pro`: the turn succeeds with no 400, and reasoning still surfaces via `reasoning_content`.

Happy to add a `CHANGELOG.md` entry in your preferred format if you'd like.


gateway/app/src/main/kotlin/splice/app/Daemon.kt
```diff
@@ -315,9 +315,14 @@ public class Daemon(
...
-                    ChatQuirks(providerTag = key, sessionCacheKeyPrefix = label, emitUsageInStream = true)
+                    ChatQuirks(
+                        providerTag = key,
+                        sessionCacheKeyPrefix = label,
+                        emitUsageInStream = true,
+                        emitReasoningEffort = providerCfg.quirks.emitReasoning,
+                    )
                 } else {
-                    ChatQuirks(providerTag = key)
+                    ChatQuirks(providerTag = key, emitReasoningEffort = providerCfg.quirks.emitReasoning)
```
marcospaulo, line 352: Wiring verified: this path previously ignored `providerCfg.quirks` and leaned on the `ChatQuirks` default, so default-`true` keeps every existing config's requests byte-identical.

One note, non-blocking: this seam (TOML quirk → `ChatQuirks`) is the one link the tests don't cover — `EmitReasoningQuirkTest` proves TOML → `QuirksConfig`, and `ChatRequestBuilderTest` proves `ChatQuirks(emitReasoningEffort = false)` → fields omitted, but nothing asserts the daemon copies the value across. `chatProvider` is private and the `tool_choice` quirk (#20) has the same untested shape, so this matches precedent — just flagging it as the spot a future regression would slip through silently.

gateway/core/src/main/kotlin/splice/core/topology/Topology.kt
```diff
@@ -106,6 +106,11 @@ public data class QuirksConfig(
     @SerialName("summary_field") val summaryField: Boolean = true,
     @SerialName("compact_effort") val compactEffort: String? = null,
     @SerialName("tool_choice") val toolChoice: Boolean = false,
+    /** openai-chat only: emit the `reasoning`/`reasoning_effort` request fields (default true,
+     *  matching OpenRouter/xAI). Set false for strict OpenAI-compatible vendors (e.g. Fireworks)
+     *  that reject unknown request fields; the reasoning mirror still reads `reasoning_content`
+     *  from the response. */
+    @SerialName("emit_reasoning") val emitReasoning: Boolean = true,
```
marcospaulo, line 113: Discoverability suggestion (fine as a follow-up): the openai-chat block in `config/splice.example.toml` (the `[providers.openrouter]` one) already suggests swapping in DeepSeek/Ollama/LM Studio — exactly the vendors this quirk exists for. A commented line there, e.g.

```toml
# quirks = { emit_reasoning = false }  # strict vendors (e.g. Fireworks) reject the reasoning field
```

would make this reachable without reading `QuirksConfig`.

(The `emit_reasoning` vs `emitReasoningEffort` naming asymmetry reads as intentional and better on the TOML side — it gates both request fields — so no change wanted there.)

marcospaulo (commented):
**LGTM.** Every claim in the PR body checked out against the code. Since the branch is behind `main` and no CI ran on this PR, I verified the merged state locally: clean merge onto `ca16f0f`, then `:core:check`, `:app:check` (76 tests, 0 failures, including the 2 new ones), `:dialect-openai-chat:test`, `:arch-tests:test`, and the ast-grep rules gate (scan + 31 rule tests) — all green.

Verified specifically:
- Before this PR the chat path ignored `providerCfg.quirks` entirely and relied on the `ChatQuirks.emitReasoningEffort = true` default, so wiring `providerCfg.quirks.emitReasoning` (default `true`) is behavior-preserving for every existing config, including grok-oauth.
- `ChatRequestBuilder` suppresses both `reasoning_effort` and `reasoning` when the quirk is off (already pinned by `ChatRequestBuilderTest`), and the response-side `reasoning_content` mirror is untouched.
- The new `QuirksConfig` field is inert for the Responses dialect — the overlay copies named fields only, so nothing leaks there.

The new test file matches the app-module idiom (root package, leading design comment, realistic full topology), and both cases are meaningful: default-true is the regression guard for OpenRouter/xAI, explicit-false is the Fireworks fix.

On the CHANGELOG offer: yes please — one line under `### Fixed` in the existing format would be great.

marcospaulo:
I'll approve it once you resolve this minor comments. Appreciate the contribution @calube

marcospaulo:
Closing: issue #21 is now fixed on main via the `quirks.reasoning_effort` nullable-overlay knob (PR #46, commit 40d856b), which follows the same TOML-overlay idiom the other quirks standardized on this week (`reasoning_cache`, `tool_choice`). This PR's `emit_reasoning` approach was sound and the review feedback on it fed the final shape — thank you. If the landed knob misses a case you needed, please open an issue and we'll extend it.

PR #46, opened 2026-07-24
feat(responses): gateway-held reasoning cache — codex-parity tool-loop continuity

### LINKED ISSUE #21: openai-chat dialect always sends reasoning/reasoning_effort — strict OpenAI-compatible vendors (Fireworks) 400

### LINKED ISSUE #43: webui dependency tree is incoherent — frozen by npm ci, blocks plugin-react 6
Found while resolving #29 (@vitejs/plugin-react 4.7.0 → 6.0.4). The bump itself is compatible (plugin-react 6.0.4 peers `vite: ^8.0.0`, webui declares vite ^8.1.5) but cannot land because the lockfile carries pre-existing debt that only survives because CI runs `npm ci` (verbatim lockfile install, never re-resolved):

1. **Phantom root `vite@6.4.3`**: `node_modules/vite` is 6.4.3 while `webui/node_modules/vite` is 8.1.5. Nothing declares vite 6 — pure lockfile inertia. Today every plugin types against the phantom consistently, so tsc passes; plugin-react 6 must type against vite 8, so the `plugins: [react(), viteSingleFile()]` array mixes vite-8-typed and vite-6-typed `Plugin` and fails TS2769 (the #29 gate failure). `vite-plugin-singlefile` 2.3.3 already peers `^8.0.0`, so the phantom is removable.
2. **typescript 7 vs typescript-eslint**: webui pins `typescript ^7.0.2` but `typescript-eslint 8.65.0` peers `>=4.8.4 <6.1.0`. Any fresh `npm install` ERESOLVEs (verified 2026-07-24); the current tree only exists because it was never re-derived. Note #8 (the typescript 7.0.2 bump PR) was closed, yet the manifest carries ^7.0.2 — worth deciding which is intended.

**Fix plan** (one sitting, ~30 min):
- Decide typescript: either pin back to <6.1 or wait for/bump a typescript-eslint that peers TS 7.
- Delete the lockfile, fresh `npm install` (no legacy-peer-deps), verify a SINGLE vite@8.1.5 in the tree.
- Re-apply the plugin-react 6 bump (it resolves trivially then), rebuild the committed webui dist (gate byte-check).

Until then #29 stays closed and dependabot ignores plugin-react majors (companion PR).

marcospaulo:
Fixes the gpt-5.6 'confused as fuck' agent behavior through the gateway: with `replay_reasoning = false` splice dropped every encrypted reasoning item between requests, so the model lost the plan motivating each tool call — repeated tool calls, duplicated reasoning, wrong-tool selection. codex-rs never has this because it replays reasoning items on every round-trip (`store:false` stateless full replay, verified at source: `client.rs:888/:915`) while displaying summaries on an independent channel.

This PR holds the encrypted envelopes **gateway-side** (codex holds them process-side) — no `redacted_thinking` storage channel through Claude Code, so none of the empty-spinner walls that got replay turned off in the first place.

**Mechanism**: capture at the tool-use terminal (real call ids only) → bounded+TTL cache (256 turns/64MB/30min, read-not-consume) → inject ONCE immediately before the turn's FIRST `function_call` on the next request (rs_-id at-most-once across both injection paths) → `invalid_encrypted_content` strips-and-retries once via a new one-shot `UpstreamClient.amendBodyOnFailure` seam. Cache miss/eviction/restart = today's behavior exactly. Fresh user turns get no injection (server `current_turn` default; deliberate divergence from codex's all-turns retention). `quirks = { reasoning_cache = false }` escape hatch; retention documented in SECURITY.md.

**Receipts**: eli source verification of openai/codex (full report on the campaign ledger `dev/campaigns/reasoning-cache.toml`), including the constraints this design honors: reasoning-must-precede-its-function_call (400 both directions), verbatim rs_ ids at most once, parallel_tool_calls already forced off on the 5.6 lite path.

**Live probe (RC-6)**: `checks/e2e/reasoning-cache-probe.sh` — two isolated daemon passes against a recording mock upstream. Wire receipts: the follow-up request carries the turn's envelope exactly once, immediately before the FIRST of two parallel `function_call`s; round 3 carries both turns' envelopes each in-position; a 400 `invalid_encrypted_content` retries once byte-identical minus the reasoning items with the client seeing a clean turn; `reasoning_cache = false` leaves the follow-up amnesiac on the wire (the recorded before-state). Deterministic — no credentials, no provider bill. 15 new walls + probe green; full `gradlew check` + ast-grep clean. Campaign ledger `dev/campaigns/reasoning-cache.toml` rides this PR with per-item receipts.

---

**Also on this branch (commit `4f40d86`)**: the post-review bug sweep BS-1..BS-6 (see the PR comment for the full map) — content-filter honesty in the responses dialect, typed `PersistFailed` on token-rotation write failures, deadline-gated stream reissue, the bounded SIGTERM-safe daemon shutdown + restart port guard, the `reasoning_effort` quirks knob, and the webui dependency-tree repair.

Closes #21
Closes #43


marcospaulo (commented):
Review of the gateway-held reasoning cache (capture at stream end, in-position reinjection on tool round-trips, 400-staleness strip-and-retry, quirks off-switch, live e2e probe). Pipeline: deterministic ast-grep scan, five review lenses plus a state-seam flow map (all evidence-gated), then adversarial filtering. 31 raw findings, 14 survived (0 blocker, 2 concern, 11 suggestion, 1 nit), 17 dissolved with recorded reasons.

The two concerns, in rank order: a single `invalid_encrypted_content` 400 evicts every cached round in the re-sent history rather than just the stale one (defeating the read-not-consume design for deep loops), and the RC-4 amend-retry seam ships with zero unit tests plus a latent boundary bug (an amend computed at the last permitted attempt is never sent). Neither regresses below the pre-PR status quo, which is why they are concerns rather than blockers.

Checked clean, explicitly: TOML overlay chaining at all 4 provider construction sites; probe assertions verified non-vacuous; SECURITY.md retention claims (TTL, bounds, opaque-ciphertext-only, memory-only) verified accurate against the code; no envelope logging anywhere; no secrets in any materialized file; the probe script's isolation from a live daemon; the PR's "15 walls" count verified exact.

Mechanical scan: run (93 rules: bundled pack + this repo's own .rules). All 16 hits on added lines dissolved as false positives for this codebase, 14 of them generic starter-pack style rules contradicting the repo's established idiom (its own 31-rule wall pack and detekt are green on this head). One decay signal worth noting: the bundled `no-unsafe-cast` rule fired on `el as? JsonObject`, which is already the safe form the rule recommends.

Cross-cutting observations with no single anchorable line: the per-provider cache instance is process-global, so the 256-entry/64MB budget is shared across every concurrent session on a head (relevant to multi-seat use); fold/reanchor continuation rounds interact with capture through seams in ResponsesReanchorController/FoldController (out of this PR's diff, worth a look when touching those); the ledger's ATTEST lines are identical boilerplate across all six items.

One file, `ReasoningCache.kt`, carries four of the fourteen comments. That concentration reflects it being the PR's new core data structure (every lens and the flow map converged there), not the quality of the file.

Optional follow-up, on request: a draft ast-grep rule (`kt-no-mutable-collection-default-in-data-class`) for the mutable-collection-default-in-data-class copy() aliasing pattern this review confirmed.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ReasoningCache.kt
```diff
@@ -0,0 +1,113 @@
...
+    var dropped = 0
+    val kept = buildJsonArray {
+        base.input.forEach { el ->
+            val item = el as? JsonObject
+            if (item != null && item["type"].str() == "reasoning") {
+                dropped++
+            } else {
+                add(el)
+                if (item != null && item["type"].str() == "function_call") {
+                    item["call_id"].str()?.let { cache.evictByToolId(it) }
```
marcospaulo: **[concern]** The RC-4 recovery correctly degrades to amnesia rather than failing the turn, and stripping every reasoning item on a 400 is the documented choice (ReasoningCache.kt:88). The eviction side has a wider blast radius than the strip needs, though. Because `store:false` re-sends the full history every turn, `stripStaleReasoning` walks every `function_call` in that history and calls `cache.evictByToolId(call_id)` on each (ReasoningCache.kt:104-105). One entry is shared by a whole round's ids, so a single `invalid_encrypted_content` 400 on one stale round evicts every other still-fresh round's entry too, and the conversation loses cache continuity for the rest of its life, not just for the turn that 400'd. The e2e probe misses this because scenario B only ever has one round in flight when the 400 fires.

If the intent is to stop re-offering only the round(s) whose envelopes were actually rejected, scope the eviction to the dropped reasoning items rather than to every retained `function_call`. Roughly: while walking `base.input`, record the round's `call_id`s as you drop each reasoning item, then evict only those:

```kotlin
val staleCallIds = mutableListOf<String>()
// ...when a reasoning item is dropped, resolve the round it belongs to and collect its call_ids...
staleCallIds.forEach { cache.evictByToolId(it) }
```

If wiping the whole conversation's cache on any staleness event is deliberate (the simplest safe fallback), a one-line note at line 104 saying so would keep the next reader from reading it as a bug.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamClient.kt
```diff
@@ -132,10 +144,25 @@ public class UpstreamClient(
...
+        // RC-4: the one-shot body-amendment budget (a deterministic 400 amended twice is a loop).
+        var amendedOnce: Boolean = false
+
+        /** RC-4: the one-shot amendment decision — lives here because it is pure retry-state
+         *  bookkeeping (budget + attempt spend); the content policy itself is the caller's hook. */
+        fun amendStep(ctx: PostContext, outcome: RetryOutcome.Failed, bodyJson: String): LoopStep.Amend? {
+            if (amendedOnce) return null
+            val amended = ctx.amendBodyOnFailure(outcome.status, outcome.text, bodyJson) ?: return null
+            amendedOnce = true
+            attempt += 1
```
marcospaulo: **[concern]** The one-shot amend path is a clean way to slot content recovery into the retry loop. Two behaviors it introduces have no unit coverage anywhere in the PR, though (the diff touches no `UpstreamClient*Test.kt`), and both are easy to get wrong:

1. The `amendedOnce` latch (UpstreamClient.kt:154): a second content-rejecting failure after the first amend must fall through to normal retry/give-up, never amend twice.
2. The attempt spend at line 157. `amendStep` does `attempt += 1`, and the outer loop is `while (state.attempt < maxRetries)` (line 121). If the amend fires when `state.attempt == maxRetries - 1`, the increment trips the loop guard and the loop exits via `giveUp(state.lastErr)` without ever sending the amended body, quietly dropping the "retries immediately" guarantee and giving up on the pre-amend error.

The e2e probe can't reach either case (scenario B produces exactly one 400 then a 200, with retry headroom). A focused `UpstreamClientAmendTest.kt` mirroring the existing `UpstreamClient*Test.kt` shape would pin both:

```kotlin
// latch:  amend fires once; a second content-reject hits normal retry/give-up
// budget: amend invoked at attempt == maxRetries - 1 -> assert the amended body IS sent
//         (or, if by design it isn't, assert giveUp and document the boundary)
```

Worth deciding explicitly whether the near-budget case should guarantee the amended send. Right now the behavior is whatever the loop math happens to produce.

dev/campaigns/reasoning-cache.toml
```diff
@@ -0,0 +1,106 @@
...
+# [2026-07-24] ATTEST: scoped diffstat: (no scoped file changes); sidecar: none; off-scope: none
+
+[[items]]
+id = "RC-2"
+phase = "P1-capture"
+title = "ReasoningCache: gateway-held bounded store keyed (conversationKey, turnKey) — conversation identity via the existing CacheKeyStrategy machinery (session-id / first-message-hash), NEVER the bare tool_use id (cross-conversation bleed, eli risk 8). Bounds: envelope-count + byte ceiling sized for deepest concurrent turns; TTL well above max turn wall-time (watchdog totalCap) and turn-AWARE: never evict a turn still receiving tool_results (eli risk 4 — mid-loop eviction reproduces the amnesia symptom intermittently). Consume-on-injection with grace re-read for retried requests. Single-process only; sticky per-head routing is the deployment reality (one daemon), note the restart-loses-cache behavior explicitly (eli risk 5) — a restart degrades to status quo, never errors"
+files = ["gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ReasoningCache.kt", "gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt"]
+status = "done"
+verify = "cd gateway && ./gradlew :gateway:test"
```
marcospaulo: **[suggestion]** RC-2's `verify` is `cd gateway && ./gradlew :gateway:test`, but the fenced files live under `gateway/dialect-openai-responses/`, and both the FENCE-EDITED note (line 51) and the landed note (line 55, ":dialect-openai-responses:test green") confirm the move out of the `gateway` module. `:gateway:test` runs only the `gateway` module's own source set, so re-running RC-2's literal verify never compiles or exercises `ReasoningCacheTest.kt` and comes back green without touching the code it gates. The real repo `check` does cover the dialect module, so nothing is actually unverified, but the recorded per-item command is a stale artifact of the rename and would mislead anyone reproducing the "done" claim. One-line fix to match the fence and the note:

```toml
verify = "cd gateway && ./gradlew :dialect-openai-responses:test"
```

dev/campaigns/reasoning-cache.toml
```diff
@@ -0,0 +1,106 @@
...
+# [2026-07-24] landed: BuildOptions.reasoningLookup (+per-request injectedReasoningIds dedup set), addReasoningOnce as the SINGLE seam for both cache and legacy replay paths (id-less items pass through — real codec always carries id), injection in appendToolUse before first fc; 4 walls in ReasoningInjectionTest incl. byte-identical-miss and dual-path dedup. verify green
+# [2026-07-24] ATTEST: scoped diffstat: (no scoped file changes); sidecar: none; off-scope: none
+
+[[items]]
+id = "RC-4"
+phase = "P2-inject"
+title = "Staleness retry: an upstream 400 invalid_encrypted_content (encrypted envelopes expire server-side, eli constraint 4 + sub2api #3138 precedent) triggers ONE re-send with the injected reasoning items stripped — degrade to per-item amnesia, never fail the turn on cache contents. Wire into UpstreamClient's existing pre-handoff retry classification (it is a pre-stream 400, inside the existing retry window); evict the offending cache entry on the way"
+files = ["gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamClient.kt", "gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt"]
```
marcospaulo: **[suggestion]** RC-4's `files` fence lists only `UpstreamClient.kt` and `ResponsesProvider.kt`, but the PR also changes `gateway/provider-spi/src/main/kotlin/splice/spi/Provider.kt` (the new `amendBodyOnFailure` default hook, doc-tagged "RC-4") and `gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt` (the `amendBodyOnFailure = provider::amendBodyOnFailure` wiring). RC-4's own landed note credits both ("Provider SPI default-null hook, TurnDriver pass-through"), yet neither file appears in any item's fence across RC-1..RC-6. An auditor reconciling diff hunks to owning items reads those two hunks as undeclared scope. Both are small single-hunk changes, so adding them to RC-4's `files` is a minimal edit:

```toml
files = [
  "gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamClient.kt",
  "gateway/provider-spi/src/main/kotlin/splice/spi/Provider.kt",
  "gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt",
  "gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt",
]
```

dev/campaigns/reasoning-cache.toml
```diff
@@ -0,0 +1,106 @@
...
+# [2026-07-24] ATTEST: scoped diffstat: (no scoped file changes); sidecar: none; off-scope: none
+
+[[items]]
+id = "RC-5"
+phase = "P3-policy"
+title = "Policy, knob, docs: cache ON by default for the responses dialect (codex head), quirks = { reasoning_cache = false } escape hatch in QuirksConfig + example config; fresh user turns get NO injection (deliberate divergence from codex's physical all-turns retention — safer, matches the server current_turn default, honors the 07-20 thins-fresh-reasoning note in its only plausible scope; eli Q5); replay_reasoning documented as the legacy client-round-trip mode; RETENTION note in PRODUCT.md/SECURITY.md: gateway holds opaque encrypted envelopes, bounded, TTL'd, never plaintext"
+files = ["gateway/core/src/main/kotlin/splice/core/topology/Topology.kt", "gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt", "gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesRequestBuilder.kt", "gateway/app/src/main/kotlin/splice/app/Daemon.kt", "config/splice.example.toml", "PRODUCT.md", "SECURITY.md"]
```
marcospaulo: **[suggestion]** Two `files` fences don't reconcile with the actual 16-file diff, which weakens the fence-to-hunk map the ledger exists to provide:

1. RC-5 (line 86) lists `PRODUCT.md`, but the PR never touches it. The only diff hunk for the RETENTION note is `SECURITY.md` (+8 lines), RC-5's own landed note (line 92) says "example-config + SECURITY.md retention note" and never mentions PRODUCT.md, and the FENCE-EDITED note (line 91) carried PRODUCT.md through unchanged. So it's an undelivered entry, not a rename artifact.
2. RC-6 (line 99) lists `gateway/gateway/src/test/kotlin/ReasoningCacheTest.kt`, the pre-rename path. RC-2 already relocated that basename to `gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt`, which is the real added test. RC-6's landed note only mentions the probe script, so this reads as a stale copy-paste.

Fix: drop `PRODUCT.md` from RC-5 (or add a note that the doc was deferred), and remove the stale test path from RC-6 (it really only owns `checks/e2e/reasoning-cache-probe.sh`). Both keep the ledger reconstructable against the diff it ships beside.

gateway/core/src/main/kotlin/splice/core/topology/Topology.kt
```diff
@@ -106,6 +106,9 @@ public data class QuirksConfig(
     @SerialName("summary_field") val summaryField: Boolean = true,
     @SerialName("compact_effort") val compactEffort: String? = null,
     @SerialName("tool_choice") val toolChoice: Boolean = false,
+    /** openai-responses only: the gateway-held reasoning cache for tool round-trips (RC-5,
+     *  2026-07-24). Default on; false restores the pre-cache behavior (per-tool-result amnesia). */
+    @SerialName("reasoning_cache") val reasoningCache: Boolean = true,
```
marcospaulo: **[suggestion]** `reasoning_cache` is wired end to end (Topology.kt:111 -> Daemon `.withReasoningCacheToml` -> ResponsesQuirks), but there's no round-trip parse test proving a real TOML value reaches the field. This repo already burned on exactly this class: the `withToml` audit on 2026-07-18 found five of seven quirks hard-coded and non-functional, which is why `ExampleConfigTest.kt` asserts each quirk deserializes (`assertEquals("session-id", xai.quirks.cacheKey)`). The new field skips that pattern. `config/splice.example.toml` only gained an explanatory comment (lines 30, 33), no actual `reasoning_cache = ...` key in either `quirks = { ... }` table, and the e2e probe writes its own synthetic TOML rather than exercising real deserialization.

Adding the key plus one assertion closes it in the same shape as its siblings:

```toml
# config/splice.example.toml, in one provider's quirks table
quirks = { ..., reasoning_cache = false }
```
```kotlin
// ExampleConfigTest.kt
assertEquals(false, head.quirks.reasoningCache)
```

That proves `reasoning_cache = false` actually reaches `QuirksConfig.reasoningCache` and then `ResponsesQuirks.reasoningCache`, which is the knob's whole contract.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ReasoningCache.kt
```diff
@@ -0,0 +1,113 @@
...
+    private val clock: () -> Long = System::currentTimeMillis,
+) {
+
+    private data class Entry(val toolIds: List<String>, val envelopes: List<String>, val bytes: Long, val at: Long)
+
+    // Insertion-ordered so eviction drops the OLDEST turn first — a deep in-flight tool loop's
+    // recent entries survive bound pressure (eli risk 4: mid-loop eviction reproduces the
+    // amnesia symptom intermittently; oldest-first plus a TTL far above turn wall-time avoids it).
+    private val entries = LinkedHashMap<String, Entry>()
+    private val byToolId = HashMap<String, String>()
```
marcospaulo: **[suggestion]** The cache keys on the bare upstream `call_id` (`byToolId`, ReasoningCache.kt:34), with one instance shared per provider/head (ResponsesProvider.kt:130). RC-2's own spec asked for `(conversationKey, turnKey)` keying via `CacheKeyStrategy` and called out the bare id explicitly: "NEVER the bare tool_use id (cross-conversation bleed, eli risk 8)." The shipped `Entry` carries no conversation or session field, so the isolation the campaign designed in isn't there.

In practice OpenAI `call_id`s are high-entropy and unique per response, so a real collision is unlikely, and a genuinely mismatched envelope should upstream-400 as `invalid_encrypted_content` and self-heal through RC-4. That's incidental protection, not the designed one: two conversations sharing a head that ever echo the same `call_id` (id reuse, a buggy or compromised client head, a future mock with lower id entropy) would reinject one conversation's reasoning into another's outgoing request before any 400 can catch it.

Threading the existing conversation-scoping key into `put`/`lookup`/`evictByToolId`, so `entries`/`byToolId` partition per conversation, restores the eli-risk-8 defense the ledger intended:

```kotlin
fun put(conversationKey: String, toolIds: List<String>, envelopes: List<String>)
fun lookup(conversationKey: String, toolId: String): List<String>?
```

If bare-id keying was a deliberate simplification given the entropy argument, a note on `byToolId` recording that decision (and that it diverges from RC-2) would close the gap between spec and code.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesRequestBuilder.kt
```diff
@@ -138,6 +146,15 @@ public data class BuildOptions(
...
     val decodeReasoningEnvelope: (String) -> JsonObject?,
+    /** RC-3 (reasoning-cache 2026-07-24): the gateway-held cache lookup — tool_use id → the
+     *  ordered envelopes of the turn that emitted it. Null = miss = today's behavior exactly.
+     *  Wired by the provider; the default keeps unwired builds byte-identical. */
+    val reasoningLookup: (String) -> List<String>? = { null },
+    /** Per-REQUEST rs_-id dedup across BOTH injection paths (cache + legacy client replay):
+     *  upstream 400s a duplicated reasoning id, and one turn's entry is shared by all its
+     *  parallel tool_use blocks — first render injects, the rest skip (inject-once law).
+     *  NB: BuildOptions.copy() would SHARE this set — construct fresh per request, never copy. */
+    val injectedReasoningIds: MutableSet<String> = mutableSetOf(),
```
marcospaulo: **[suggestion]** The inject-once dedup set is the right idea, and the KDoc already flags the copy() hazard (line 156). Right now that hazard is guarded only by a comment. `injectedReasoningIds` is a `MutableSet` in a public `data class`, so `BuildOptions.copy()` aliases the reference rather than deep-copying it, and defaults aren't re-evaluated on copy. No call site copies `BuildOptions` today, so nothing breaks yet, but the first `opts.copy(compact = true)` added later would silently share the dedup set between the two instances, letting one suppress the other's `rs_` injections and producing exactly the orphaned-reasoning request RC-4 exists to recover from, with no exception to trace it.

Since nothing overrides `injectedReasoningIds` anywhere (only `reasoningLookup` is ever customized), the cleanest fix removes the landmine structurally instead of relying on call-site discipline: thread the set as a plain parameter at the actual build boundary, or drop `data` from `BuildOptions` (nothing appears to use its `equals`/`hashCode`/`copy`):

```kotlin
// option: not part of the copyable surface at all
public class BuildOptions( ... )   // plain class, no synthesized copy()
```

Either one makes the invariant enforced rather than documented.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ReasoningCache.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /** Drop the turn containing [toolId] — used when upstream rejects its envelopes as stale. */
+    fun evictByToolId(toolId: String) {
+        synchronized(lock) { byToolId[toolId]?.let { removeLocked(it) } }
+    }
+
+    private fun sweepLocked() {
+        val cutoff = clock() - ttlMs
+        val expired = entries.entries.takeWhile { it.value.at < cutoff }.map { it.key }
```
marcospaulo, line 93: **[suggestion]** `sweepLocked` stops at the first entry with `at >= cutoff` via `takeWhile` (ReasoningCache.kt:69), which is correct only while insertion order matches ascending timestamps. The cache clock is `System::currentTimeMillis` (line 25, a wall clock). This PR's own `UpstreamClient` deliberately uses `MonoClock::nowMs` for exactly this reason (line 68: "a wall-clock jump must not abort a healthy retry loop"). If the wall clock steps backward between two `put`s (NTP correction, manual adjustment), a later-inserted entry can carry an earlier `at` than one before it, `takeWhile` then stops early, and the later-inserted-but-expired entry is never TTL-swept (it leaves only via the hard 256-entry / 64MB bound). Not a crash (status quo holds), but it quietly breaks the documented 30-minute TTL on long-running daemon hosts, where clock corrections do happen. Aligning the cache with the same monotonic clock the rest of this PR standardizes on removes the assumption:

```kotlin
private val clock: () -> Long = MonoClock::nowMs,
```

With a monotonic clock, insertion order and `at` order can't diverge, so the `takeWhile` sweep is sound.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
```diff
@@ -112,6 +125,17 @@ public abstract class ResponsesProvider(
...
     }
 
+    // RC-2/RC-4: gateway-held reasoning continuity for tool round-trips (codex parity). One
+    // cache per provider instance; capture and lookup wire in via buildTurn/streamTranslator.
+    private val reasoningCache: ReasoningCache = ReasoningCache()
+
+    /** RC-4: a 400 rejecting stale encrypted reasoning strips the injected items and retries
+     *  once (NEVER-BELOW-STATUS-QUO law); every other failure keeps the plain retry plan. */
+    final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? {
+        if (status != HTTP_BAD_REQUEST || !responseText.contains("invalid_encrypted_content")) return null
```
marcospaulo: **[suggestion]** `amendBodyOnFailure` gates on `responseText.contains("invalid_encrypted_content")` (ResponsesProvider.kt:135), an exact case-sensitive match, but it has to win a race against the pre-existing classifier `isEncryptedContentError` (UpstreamClient.kt:650-651), which matches the same failure class far more broadly: `contains("encrypted_content", ignoreCase = true)`. `amendStep` runs before `planStep` (UpstreamClient.kt:217), so the narrow gate goes first, but any 400 whose wording mentions `encrypted_content` without the literal `invalid_encrypted_content` (different casing, a reworded message, a sibling error code) fails the amend check, falls through to `planRetry`, and hits `isEncryptedContentError` -> GIVE_UP with no retry at all. Since this PR requests `encrypted_content` and injects reasoning much more aggressively, it also raises how often this 400 class can fire, so the narrow gate is a single point of failure for the recovery path the feature adds, and the e2e probe (which uses the exact matched string) can't surface a wording drift.

Sharing one classifier keeps both gates in lockstep:

```kotlin
// reuse the same predicate both sides key off
if (status != HTTP_BAD_REQUEST || !isEncryptedContentError(status, responseText)) return null
```

gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        val c = ReasoningCache(clock = { 0L })
+        c.put(listOf("call_a", "call_b"), listOf("e1"))
+        c.evictByToolId("call_b")
+        assertNull(c.lookup("call_a"))
+    }
+
+    @Test
+    fun `empty puts are ignored`() {
+        val c = ReasoningCache(clock = { 0L })
+        c.put(emptyList(), listOf("e1"))
```
marcospaulo: **[suggestion]** `empty puts are ignored` calls `put` twice (empty toolIds, then empty envelopes) but funnels both through one `assertNull(c.lookup("call_a"))` (ReasoningCacheTest.kt:63-65). That assertion can only catch the `envelopes.isEmpty()` half of the guard. If the `toolIds.isEmpty()` half (ReasoningCache.kt:40) were removed, `put(emptyList(), listOf("e1"))` would create an orphan entry: `toolIds.forEach { byToolId[it] = key }` iterates zero times, so nothing links to `"call_a"` and the lookup is still null, test still green, even though the entry now occupies a `maxEntries`/`maxTotalBytes` slot and can evict a real turn under bound pressure. Split the cases and give the toolIds-empty case an observable side effect of the guard firing:

```kotlin
@Test fun `empty toolIds are ignored`() {
    val c = ReasoningCache(maxEntries = 1, clock = { 0L })
    c.put(emptyList(), listOf("e1"))            // must NOT consume the single slot
    c.put(listOf("call_a"), listOf("e2"))
    assertEquals(listOf("e2"), c.lookup("call_a"))  // fails if an orphan evicted it
}

@Test fun `empty envelopes are ignored`() {
    val c = ReasoningCache(clock = { 0L })
    c.put(listOf("call_a"), emptyList())
    assertNull(c.lookup("call_a"))
}
```

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
```diff
@@ -99,7 +108,11 @@ public abstract class ResponsesProvider(
                 // could consume them: fold replay (Success side) OR mid-stream re-anchor salvage
                 // (Failure side) — i.e. every non-compact responses turn since re-anchor landed
                 // (2026-07-24). Compact turns keep the collection off.
-                collectReasoningEnvelopes = foldController(meta) != null || reanchorController(meta) != null,
+                collectReasoningEnvelopes = foldController(meta) != null || reanchorController(meta) != null ||
```
marcospaulo, line 117: **[suggestion]** Both cache gates add a `!compact`/`!meta.compact` conjunct on top of `quirks.reasoningCache` (capture at ResponsesProvider.kt:114, collect at 111-112, and the lookup gate on the builder side), but nothing tests the `!compact` term. `TurnReasoningCaptureTest` covers the generic `onTurnReasoning` callback, `ReasoningInjectionTest` covers the builder's use of a lookup, and the e2e probe covers only the ON/OFF toggle, none of them a compact turn. A regression dropping the `!compact` conjuncts (so compaction turns start reading and writing the cache) would pass every check in this PR. Either add a compact pass to the probe, or a narrow provider-level test:

```kotlin
// with quirks.reasoningCache = true AND meta.compact = true:
// assert reasoningLookup returns null and onTurnReasoning is a no-op
```

Low probability, but it's a silent gate on a real branch, so cheap to pin.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesRequestBuilderTest.kt
```diff
@@ -397,3 +397,96 @@ class ResponsesRequestBuilderTest {
...
+        val req = build(body, options = cacheOpts(lookup = { listOf("env1") }))
+        val input = items(req)
+        val reasonings = input.filter { it["type"]?.jsonPrimitive?.content == "reasoning" }
+        assertEquals(1, reasonings.size, "inject-once per turn")
+        val firstFc = input.indexOfFirst { it["type"]?.jsonPrimitive?.content == "function_call" }
+        assertEquals("reasoning", input[firstFc - 1]["type"]?.jsonPrimitive?.content)
+    }
+
+    @Test
+    fun `a cache miss is byte-identical to an unwired build`() {
```
marcospaulo: **[suggestion]** The intent of `a cache miss is byte-identical to an unwired build` is a real backward-compat guarantee worth pinning, but as written both sides are the same no-op: `cacheOpts(lookup = { null })` on the left, and `cacheOpts()` on the right whose `lookup` parameter also defaults to `{ null }` (ResponsesRequestBuilderTest.kt:406). Both lambdas unconditionally return null, and `appendToolUse`'s `opts.reasoningLookup(block.id)?.forEach { ... }` is a complete no-op on null regardless of which null-returning lambda supplied it, so no cache-miss regression could ever make these two builds differ. The generic "purity" test already covers default-parameter resolution.

To actually test the compat claim (a caller unaware of the cache gets the same bytes as a cache-aware caller that sees a miss), compare against the pre-cache `opts()` helper, which never mentions `reasoningLookup`:

```kotlin
val unwired = build(TOOL_TURN_BODY, options = opts(/* pre-cache helper, no reasoningLookup */))
val miss    = build(TOOL_TURN_BODY, options = cacheOpts(lookup = { null }))
assertEquals(unwired.toString(), miss.toString())
```

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ReasoningCache.kt
```diff
@@ -0,0 +1,113 @@
...
+    fun put(toolIds: List<String>, envelopes: List<String>) {
+        if (toolIds.isEmpty() || envelopes.isEmpty()) return
+        val bytes = envelopes.sumOf { it.length.toLong() }
+        synchronized(lock) {
+            sweepLocked()
+            val key = "t${seq++}"
+            entries[key] = Entry(toolIds, envelopes, bytes, clock())
+            toolIds.forEach { byToolId[it] = key }
+            totalBytes += bytes
+            while (entries.size > maxEntries || totalBytes > maxTotalBytes) {
```
marcospaulo, line 66: **[nit]** `byte ceiling evicts oldest until under bound` sizes its scenario so exactly one eviction restores the bound (10-byte ceiling, 5-byte entries), so the second-and-later iteration of the `while` at ReasoningCache.kt:48 is never exercised. The per-iteration logic (`removeLocked`) is identical each pass, so risk is low, but it is the literal untested branch of the loop. A case whose single `put` pushes `totalBytes` far enough over the ceiling to require evicting two or more prior entries in one call, asserting both are gone, would cover it:

```kotlin
val c = ReasoningCache(maxTotalBytes = 10, clock = { 0L })
c.put(listOf("call_1"), listOf("aaaaa"))
c.put(listOf("call_2"), listOf("bbbbb"))
c.put(listOf("call_3"), listOf("ccccccccccc"))  // 11 bytes -> must evict call_1 AND call_2
assertNull(c.lookup("call_1"))
assertNull(c.lookup("call_2"))
assertEquals(listOf("ccccccccccc"), c.lookup("call_3"))
```

marcospaulo:
Review addressed in 8d0eded, all 14 comments:

- **c1 scoped eviction**: stripStaleReasoning now evicts only the rounds whose reasoning items were dropped (the unbroken run of function_calls after each); wall added with an orphan-round case.
- **c2 amend boundary + coverage**: the amend no longer spends an attempt (the amendedOnce latch is the whole budget), so the amended resend is guaranteed at the budget boundary; new UpstreamClientAmendTest pins the boundary, the one-shot latch, and the byte swap.
- **c3/c4/c5 ledger**: RC-2 verify targets :dialect-openai-responses:test; RC-4 fences Provider.kt + TurnDriver.kt; RC-5 drops PRODUCT.md; RC-6 drops the stale pre-rename path.
- **c6 round-trip proof**: reasoning_cache is now a nullable overlay knob (absent keeps the provider default), the example config carries a real non-default value (xai: false, since xai returns no envelopes; grok's code default is now off too), ExampleConfigTest asserts it, and withReasoningCacheToml has overlay walls.
- **c7 conversation scoping**: implemented per the RC-2 spec — entries key on (first-message hash, call_id) via TurnMeta.conversationKey; lookups never cross conversations; eviction stays deliberately bare-id (over-evict = miss, never a wrong injection); SECURITY.md documents it.
- **c8 copy() aliasing**: injectedReasoningIds moved to a data-class BODY property — copy() re-initializes it fresh; the hazard is gone structurally (repo forbids suppressing LongParameterList, so de-data-classing was not the route).
- **c9 MonoClock**: done.
- **c10 shared predicate**: the amend gate keys off UpstreamClient.isEncryptedContentError (now public).
- **c11/c12/c13/c14 walls**: empty-put guards split with an observable slot assertion; reasoningCacheActive(quirks, compact) is the one tested gate all four touch points route through; the byte-identical wall compares against a genuinely pre-cache BuildOptions; multi-eviction loop covered.

Full gradlew check, 31 ast-grep walls, and the e2e probe are green on the fix commit.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
```diff
@@ -54,13 +54,22 @@ public abstract class ResponsesProvider(
...
+                includeEncryptedReasoning = RequestEncryptedReasoning(
+                    (showOn || quirks.reasoningCache) && !compact,
+                ),
                 sessionId = sessionId,
                 decodeReasoningEnvelope = { decodeReasoningEnvelope(it) },
+                // RC-5: gateway-held reasoning continuity — the turn that emitted these tool ids
+                // left its plan in the cache; reinject it so the model resumes instead of
+                // re-deriving (codex parity; repeated-tool-call amnesia otherwise).
+                reasoningLookup = { id ->
+                    if (quirks.reasoningCache && !compact) reasoningCache.lookup(id) else null
```
marcospaulo: **[concern]** The cache lookup is nicely isolated behind `reasoningLookup`, and the correctness and security lenses converge on the same edge here. At `gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt:71`, the lookup is `reasoningCache.lookup(id)` even though `sessionId` is already available in this build path and the cache lives on the provider. If conversation A caches `call_abc` and conversation B later submits the same client-visible id before expiry, B receives A's encrypted reasoning envelope. That crosses the conversation boundary and can also corrupt the reverse index when ids collide.

Please thread the existing server-owned conversation key through both capture and lookup, then key the cache by the pair rather than by raw call id. For example:

```kotlin
data class ReasoningCacheKey(
    val conversationKey: String,
    val callId: String,
)

fun put(key: ReasoningCacheKey, envelopes: List<String>) { /* bounded store */ }
fun lookup(key: ReasoningCacheKey): List<String>? = /* lookup */

reasoningLookup = { id ->
    if (quirks.reasoningCache && !compact) {
        reasoningCache.lookup(ReasoningCacheKey(conversationKey, id))
    } else {
        null
    }
}
```

The important part is that `conversationKey` comes from the gateway's existing server-owned session or first-message identity machinery, not another client-supplied call id. A regression test with two conversation keys and the same `call_abc` should show that only the originating conversation resolves the envelope.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamClient.kt
marcospaulo: **[concern]** The one-shot `amendedOnce` guard prevents an infinite amendment loop, which is the right safety property. The correctness and flow-map lenses both find the same budget ordering issue. At `gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamClient.kt:157`, `attempt += 1` runs before `LoopStep.Amend` returns. If `maxRetries` is 1, or `invalid_encrypted_content` arrives on the final ordinary attempt, the enclosing `while (state.attempt < maxRetries)` check fails on the next iteration and the stripped body is never posted. The caller sees the original 400 even though the retry log says it is retrying once.

Please keep the amendment budget separate from the ordinary attempt counter, so scheduling the amended resend does not consume the slot that must execute it. The minimal state transition is:

```kotlin
fun amendStep(
    ctx: PostContext,
    outcome: RetryOutcome.Failed,
    bodyJson: String,
): LoopStep.Amend? {
    if (amendedOnce) return null
    val amended = ctx.amendBodyOnFailure(
        outcome.status,
        outcome.text,
        bodyJson,
    ) ?: return null

    amendedOnce = true
    ctx.onRetry("amending request body after ${outcome.status} and retrying once")
    return LoopStep.Amend(amended)
}
```

The key change is that this branch does not increment `attempt`; the amended request gets sent, and a later non-amendable failure can follow the ordinary budget path. Please pin it with a `maxRetries = 1` test that queues a 400 followed by success and asserts two POST bodies, with reasoning present only in the first.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesStreamTranslatorTest.kt
```diff
@@ -495,3 +497,72 @@ class ResponsesStreamTranslatorTest {
...
+                        """"item":{"type":"function_call","call_id":"call_abc","name":"run"}}""",
+                ),
+                ev("""{"type":"response.function_call_arguments.done","output_index":1,"arguments":"{}"}"""),
+                completed,
+            ).asFlow(),
+            RecordingSink(),
+        )
+        assertTrue(outcome is TurnOutcome.Success)
+        assertEquals(listOf("call_abc"), captured?.first)
+        assertEquals(1, captured?.second?.size)
```
marcospaulo, line 566: **[suggestion]** The test already proves that the callback fires once with the expected call id. At `gateway/dialect-openai-responses/src/test/kotlin/ResponsesStreamTranslatorTest.kt:529`, however, `assertEquals(1, captured?.second?.size)` does not verify which envelope was captured. If the translator drops or mutates `encrypted_content` but still emits one id-derived value, this stays green, and the builder tests cannot catch it because they inject handcrafted envelopes.

Please put a unique encrypted-content sentinel in the SSE fixture, make the test encoder include it, and assert the exact callback payload. For example:

```kotlin
val encryptedSentinel = "cipher-sentinel-9f2"
val expectedEnvelope = "rs_abc|$encryptedSentinel"

val translator = ResponsesStreamTranslator(
    encodeReasoningEnvelope = { item ->
        val id = item.getValue("id").jsonPrimitive.content
        val encrypted = item.getValue("encrypted_content").jsonPrimitive.content
        "$id|$encrypted"
    },
    onTurnReasoning = { ids, envelopes -> captured = ids to envelopes },
)

assertEquals(listOf("call_abc"), captured?.first)
assertEquals(listOf(expectedEnvelope), captured?.second)
```

The exact encoder can match the production envelope format. The key assertion is full-value equality, not collection size.

dev/campaigns/reasoning-cache.toml
marcospaulo: **[suggestion]** RC-2's file fence points at the right implementation and test module, and the landing note records that the dialect test task passed. The executable proof field does not preserve that result. At `dev/campaigns/reasoning-cache.toml:50`, `verify = "cd gateway && ./gradlew :gateway:test"` does not target the `dialect-openai-responses` test source set declared on line 48. A future manifest-driven verification can therefore mark RC-2 green without running `ReasoningCacheTest`.

Please make the declared proof target the owning module:

```toml
verify = "cd gateway && ./gradlew :dialect-openai-responses:test"
```

If `:gateway:test` is independently required for this item, append it rather than substituting it for the dialect task.

dev/campaigns/reasoning-cache.toml
```diff
@@ -0,0 +1,106 @@
...
+# [2026-07-24] ATTEST: scoped diffstat: (no scoped file changes); sidecar: none; off-scope: none
+
+[[items]]
+id = "RC-6"
+phase = "P4-proof"
+title = "Proof at the level that failed us before: the 2026-07-23 single-shot probe validated one tool call and missed the amnesia class entirely — the acceptance here is a LIVE multi-tool agentic turn through the claudex head asserting (a) request 2 carries the rs_ reasoning item immediately before the replayed function_call (capture upstream_req via a local echo probe or daemon debug), and (b) a before/after behavioral comparison on a scripted 3-tool task: repeated-tool-call count and duplicated-reasoning length drop. Plus full gradlew check + ast-grep walls green"
+files = ["checks/e2e/reasoning-cache-probe.sh", "gateway/gateway/src/test/kotlin/ReasoningCacheTest.kt"]
```
marcospaulo: **[suggestion]** The RC-6 proof command and landing note provide a useful deterministic receipt, but the file fence still names the pre-move test location. At `dev/campaigns/reasoning-cache.toml:99`, the entry points to `gateway/gateway/src/test/kotlin/ReasoningCacheTest.kt`, while the landed test and the RC-2 fence place it under `gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt`. Resume or audit tooling following RC-6 will inspect the wrong proof artifact.

Please update the fence to the landed path:

```toml
files = [
  "checks/e2e/reasoning-cache-probe.sh",
  "gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt",
]
```

That keeps the campaign's scope record aligned with the artifact the proof actually exercised.

gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt
```diff
@@ -0,0 +1,100 @@
...
+            """{"type":"function_call","call_id":"call_a","name":"run","arguments":"{}"},""" +
+            """{"type":"function_call_output","call_id":"call_a","output":"ok"}],""" +
+            """"store":false,"stream":true}"""
+
+    @Test
+    fun `strips reasoning items, keeps everything else in order, evicts the turn`() {
+        val cache = ReasoningCache(clock = { 0L })
+        cache.put(listOf("call_a"), listOf("stale"))
+        val amended = stripStaleReasoning(body, cache)!!
+        org.junit.jupiter.api.Assertions.assertFalse(amended.contains("\"reasoning\""))
```
marcospaulo: **[suggestion]** The current test does verify removal and relative item order, which covers the core happy path. The preservation claim is stronger than those assertions, though. At `gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt:86`, the test checks that the `reasoning` substring is absent, `call_a` remains somewhere, and `function_call` precedes `function_call_output`. A faulty transformation can still change the user instruction, function name or arguments, tool output, `store`, or `stream` while satisfying all three checks.

Please parse both documents and compare the amended request against the original with only reasoning input items removed. For example:

```kotlin
val original = Json.parseToJsonElement(body).jsonObject
val actual = Json.parseToJsonElement(amended).jsonObject
val expectedInput = original.getValue("input").jsonArray.filterNot { item ->
    item.jsonObject["type"]?.jsonPrimitive?.content == "reasoning"
}
val expected = JsonObject(
    original + ("input" to JsonArray(expectedInput)),
)

assertEquals(expected, actual)
assertNull(cache.lookup("call_a"), "stale turn evicted")
```

This makes the intended delta executable: reasoning disappears, every other field and input item remains structurally identical and in order.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesStreamTranslatorTest.kt
```diff
@@ -495,3 +497,72 @@ class ResponsesStreamTranslatorTest {
...
+                ev(
+                    """{"type":"response.output_item.added","output_index":1,""" +
+                        """"item":{"type":"function_call","call_id":"call_abc","name":"run"}}""",
+                ),
+                ev("""{"type":"response.function_call_arguments.done","output_index":1,"arguments":"{}"}"""),
+                completed,
+            ).asFlow(),
+            RecordingSink(),
+        )
+        assertTrue(outcome is TurnOutcome.Success)
```
marcospaulo, line 564: **[suggestion]** The successful capture coverage at `gateway/dialect-openai-responses/src/test/kotlin/ResponsesStreamTranslatorTest.kt:527` does a good job covering completed encrypted reasoning, synthetic-id handling, and text-only success. This medium-severity gap is the terminal negative path: if completed encrypted reasoning and a real `function_call` ID arrive before `response.failed`, or before EOF with no terminal success event, a regression that moves `onTurnReasoning` ahead of terminal classification will invoke the callback, seed `ReasoningCache`, and let the next tool-loop request resolve and reinject an envelope from a rejected or incomplete turn.

Please add both fixtures around the same capture callback, wire that callback through a cache fixture, and assert zero captures plus a cache miss:

```kotlin
@Test
fun `response failed after reasoning and real call id does not seed cache`() = runTest {
    val fixture = captureFixture()
    val outcome = fixture.translate(
        completedEncryptedReasoning(callId = "call_real_1") +
            realFunctionCall(id = "call_real_1") +
            responseFailed(),
    )

    assertTrue(outcome is TurnOutcome.Failure)
    assertTrue(fixture.captures.isEmpty())
    assertNull(fixture.cache.resolve(toolLoopRequest("call_real_1")))
}

@Test
fun `eof after reasoning and real call id does not seed cache`() = runTest {
    val fixture = captureFixture()
    val outcome = fixture.translate(
        completedEncryptedReasoning(callId = "call_real_2") +
            realFunctionCall(id = "call_real_2"),
        // Stream closes here without response.completed.
    )

    assertFalse(outcome is TurnOutcome.Success)
    assertTrue(fixture.captures.isEmpty())
    assertNull(fixture.cache.resolve(toolLoopRequest("call_real_2")))
}
```

The helper names can follow the existing event builders. The important boundary is that `onTurnReasoning` and the cache remain silent unless terminal classification is `TurnOutcome.Success`.

gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        assertEquals(listOf("e3"), c.lookup("call_3"))
+    }
+
+    @Test
+    fun `byte ceiling evicts oldest until under bound`() {
+        val c = ReasoningCache(maxTotalBytes = 10, clock = { 0L })
+        c.put(listOf("call_1"), listOf("aaaaa"))
+        c.put(listOf("call_2"), listOf("bbbbb"))
+        c.put(listOf("call_3"), listOf("ccccc"))
+        assertNull(c.lookup("call_1"))
```
marcospaulo: **[suggestion]** The test correctly proves that the oldest entry is evicted when the byte ceiling is crossed. At `gateway/dialect-openai-responses/src/test/kotlin/ReasoningCacheTest.kt:47`, it inserts `call_3`, but the assertions at lines 48 and 49 inspect only `call_1` and `call_2`. An implementation that evicts both `call_1` and the newly admitted `call_3` still passes, even though it loses the freshest tool-loop state at the capacity boundary.

Please assert the complete retained set, especially the triggering insert:

```kotlin
assertNull(c.lookup("call_1"))
assertEquals(listOf("bbbbb"), c.lookup("call_2"))
assertEquals(listOf("ccccc"), c.lookup("call_3"))
```

If the cache exposes retained bytes for tests, asserting that the total is at most 10 would also pin the bound, but the `call_3` assertion is the required regression wall.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesRequestBuilderTest.kt
```diff
@@ -397,3 +397,96 @@ class ResponsesRequestBuilderTest {
...
+        val hit = build(TOOL_TURN_BODY, options = cacheOpts(lookup = { null }))
```
marcospaulo: **[suggestion]** The byte-identity assertion is the right kind of regression wall, but the baseline is not independent yet. At `gateway/dialect-openai-responses/src/test/kotlin/ResponsesRequestBuilderTest.kt:473`, the miss uses `cacheOpts(lookup = { null })`, and line 474 builds the alleged unwired value with `cacheOpts()` again. A regression inside `cacheOpts` or the common cache path that reorders ordinary input on every miss changes both values, so `assertEquals` still passes.

Please compare the miss against a fixture or constructor path that does not reuse the helper under test. A full golden request keeps the comparison independent and protects the complete wire body:

```kotlin
@Test
fun `a cache miss is byte-identical to the pre-cache request`() {
    val actual = build(
        TOOL_TURN_BODY,
        options = cacheOpts(lookup = { null }),
    ).toString()

    assertEquals(PRE_CACHE_TOOL_TURN_REQUEST_JSON, actual)
}
```

`PRE_CACHE_TOOL_TURN_REQUEST_JSON` should contain the complete known pre-cache output for the same model and reasoning settings. An explicitly cache-disabled builder path is also fine if it is constructed independently of `cacheOpts`.

marcospaulo (commented):
Review of gateway-held encrypted reasoning continuity across tool-loop requests, including bounded retention, in-position reinjection, and one-shot recovery from `invalid_encrypted_content`.

Findings: **2 high, 7 medium**. The evidence gates dissolved 21 HIF candidates, including 16 mechanical rule hits, duplicate lens reports, two approximate anchors, and two initially plausible findings that contradicted the settled fallback contract or lacked a shipped ABI surface.

Coverage:
- Full sharded mode, 16 changed files and 1,063 added lines. All PR-head files materialized successfully, so no surviving finding has reduced confidence.
- Mechanical scan: 82 ast-grep rules run, 0 skipped, 16 added-line hits. Every hit was checked and dissolved as a false positive or non-actionable style match.
- The conditional state-flow map ran across capture, cache ownership, request injection, retry amendment, cancellation, and cleanup.
- `SECURITY.md` and `config/splice.example.toml` received mechanical-only coverage, as assigned by the risk governor.

Checked clean:
- `ReasoningCache` serializes map, byte counter, sequence, TTL sweep, and eviction operations under one monitor.
- Cancellation still propagates when no watchdog fires, and production capture currently occurs only after a successful terminal tool-use outcome.
- The live probe uses dummy credentials, loopback binding, private scratch state, and quoted path/port interpolation. No credential, command-injection, workflow-permission, dependency-pinning, or unsafe-deserialization issue survived review.
- The new public declarations are governed by repository `explicitApi()` discipline, but the dialect module has no publication or ABI-compatibility contract, so no speculative external-binary finding is included.

Unanchored coverage observations: the unit suite does not stress concurrent cache operations or an interleaved reasoning/function-call stream. Those are not posted as inline findings because this review found no concrete production failure at an added line.

Optional follow-up antibodies are available for the three confirmed recurring shapes: raw call-id-only reasoning cache lookup, retry-budget increment before an amendment resend, and a regression baseline built through the same new helper as the subject.

marcospaulo:
Commit `4f40d86` lands the bug-sweep campaign (BS-1..BS-6, ledger `dev/campaigns/bug-sweep.toml` with per-item provenance and typed verdicts) on this branch rather than a stacked PR:

- **BS-1**: `response.incomplete` with a non-`max_output_tokens` reason (content_filter et al.) now ends as a `providerReported` API_ERROR, matching the chat dialect's honesty branch; absent reason keeps the clean path.
- **BS-2**: the unguarded `writeSecure()` in codex/grok/kimi token rotation degrades to typed `PersistFailed` — a disk hiccup after a Granted rotation can no longer throw raw through SingleFlight or latch `invalid_grant` into a bricked head (red-green walls per provider).
- **BS-3**: the G5 stream-reissue backoff is deadline-gated like its sibling path.
- **BS-4**: bounded daemon shutdown (8s cooperative cap, `halt(0)` watchdog floor 2s above it), concurrent per-head stops, and a restart CLI that refuses to spawn into a still-bound control port.
- **BS-5**: `quirks.reasoning_effort` nullable overlay exposes `ChatQuirks.emitReasoningEffort` (fixes #21). NB: overlaps PR #22's `emit_reasoning` approach — one of the two should win.
- **BS-6**: plugin-react ^6 + Microsoft's TS6/TS7 side-by-side aliasing + a lockfile regenerated from a fresh resolve (fixes #43).

Provenance: 17-agent find/verify workflow — 9 sweep candidates, 3 survived adversarial refutation, 6 refuted with file:line proof; 3 known symptoms diagnosed to specs. Gates on the commit: full `gradlew check`, 31 walls, `npm ci`, e2e probe.

marcospaulo:
All 23 review threads resolved against the tree at `f36edbc`:

- The 14 threads from the first review: fixed in `8d0eded` (fix-map in the earlier comment).
- The 9 threads from the second review pass: the two concerns (conversation-scoped cache keying; the amend attempt-spend boundary) were independently confirmed and were already fixed in `8d0eded` — including the exact regression walls asked for (two-conversation same-call-id isolation; amend-at-final-attempt sends the stripped body). The RC-2 verify target and RC-6 fence corrections landed on the ledger in the same commit. The remaining three test-strengthening suggestions land in `f36edbc`: exact-envelope capture assertion via a content-aware encoder, negative capture walls (failed terminal + EOF truncation never seed the cache), and the strip wall as a structural-equality delta.

Issue #47, opened 2026-07-25
openai-responses (Codex/gpt-5.x): full client tool surface sent every request — token cost and tool-selection degradation


marcospaulo:
## Summary

On `openai-responses` heads backed by the ChatGPT Codex provider (`gpt-5.x`, notably the `gpt-5.6-*` family), splice forwards the **entire client tool list to the backend on every request**. It implements no tool-surface reduction: no `defer_loading`, no `tool_search`, and no `code_mode`.

This has two consequences: **elevated token cost** on every turn, and **degraded tool-selection behavior** on models whose official client presents them a far smaller tool surface.

A fix is in progress (see *Status* below).

## Affected

- Heads using `dialect = "openai-responses"` with the codex provider (`gpt-5.6-sol` / `-terra` / `-luna`, `gpt-5.5`, `gpt-5.4`, …).
- Impact scales with how many tools the client advertises. A plain Claude Code session sends on the order of ~15 tools; **each configured MCP server adds more**, and a heavily-MCP-configured session can advertise **60–90 tools per request**.
- Not affected today: `anthropic-passthrough` (Kimi) and `openai-chat` (Grok) heads — the mechanism described here is an OpenAI Responses feature and has no direct analogue on those wires (Kimi is being assessed separately).

## What happens

Every request to the Responses backend carries the full JSON Schema for **all** advertised tools, embedded in the `additional_tools` developer item (responses-lite shape). Nothing is deferred and nothing is searchable, so:

1. **Token cost** — the complete tool-schema payload is re-sent on every turn of every conversation. In tool-heavy sessions this can dominate request size (hundreds of KB per request).
2. **Tool-selection quality** — the model is asked to choose among dozens of tools per step. Observed symptoms include excessive sub-agent/task spawning, repeated attempts at the same tool, and stalled turns. These generally do **not** surface as API errors, so they are invisible in failure metrics and only show up as degraded agent behavior.

## Why this diverges from the reference client

OpenAI's own Codex CLI does not present `gpt-5.6` a large discrete tool surface. For the `gpt-5.6-*` models it sets `tool_mode = "code_mode_only"`, collapsing nearly all tools into a **single grammar-constrained `exec` tool** (individual tools are rendered as TypeScript declarations inside that tool's description and invoked from model-authored code). The effective choice surface is roughly **3 tools instead of dozens**. It additionally supports `defer_loading` on function tools together with the `tool_search` hosted tool.

[...]

PR #48, opened 2026-07-25
feat(responses): deferred tool surface with cache-stable gateway-answered tool_search

marcospaulo:
Closes part of #47.

## Problem

On `openai-responses` heads (codex provider, `gpt-5.6-*`), splice forwards the client's **entire** tool list on every request — measured **64–87 discrete function schemas**, median upstream request **375 KB**. Two costs:

1. **Tokens**, on every turn of every conversation.
2. **Tool-selection thrash** — the model is asked to choose among dozens of tools per step. This shows up as excessive sub-agent spawning, repeated attempts at the same tool, and stalled turns. It does *not* surface as an API error, so it is invisible in failure metrics.

OpenAI's own client does not do this: for `gpt-5.6-*` it runs `tool_mode = "code_mode_only"`, collapsing nearly all tools into a single grammar-constrained `exec` tool (~3 tools instead of dozens).

## Why not `code_mode`

Evaluated and rejected for a translation gateway. The model returns a *program*, and the tools it calls belong to the downstream client (which owns permissions, approvals, and execution). Honoring that would require suspending and resuming the program across separate HTTP round-trips; a restart, eviction, or interruption mid-program would leave the conversation unrecoverable, with side-effecting tool calls potentially re-executed on replay. That failure mode is worse than the problem being solved, and it cannot satisfy the never-below-status-quo law.

This PR ports the other sanctioned mechanism instead: **defer the long tail and answer `tool_search` at the gateway**, inside the existing multi-round turn loop (the same machinery fold/re-anchor already use — no new concurrency model, no new terminal path).

## Cache stability is the load-bearing property

This is the part worth reviewing closely. The tool payload rides at **`input[0]`** under a single `prompt_cache_key` for the whole conversation, so **anything that varies it per turn re-bills the entire cached prefix** (~94k tokens in the measured profile) — which would cost more than deferral saves and make the feature a net regression.

Three invariants keep it stable, each pinned by a test:

- The eager/deferred split is a **pure function of `(body.tools, policy)`**. No transcript input can reach it, so `input[0]` is byte-identical on every turn.
- A deferred tool used earlier in the transcript is re-declared **in position** — a `tool_search_call`/`tool_search_output` pair carrying its full schema, immediately before its `function_call` — and is **never** promoted into `input[0]`.
- The synthetic `call_id` is `sha256(toolName)`, so the injected pair is byte-identical across turns. (A counter or timestamp here would silently reintroduce the bust.)

This mirrors the reference implementation, which never injects a searched tool into `tools[]`; its own regression test pins the rule:

> `"follow-up request should rely on tool_search_output history, not tool injection"`
> — `codex-rs/core/tests/suite/search_tool.rs:782-814`

The mechanism that makes it work: `ToolSearchOutput { tools: Vec<LoadableToolSpec> }` carries the tool's **full schema into history**, so the model learns tools from history rather than from the tools array.

## Safety

- **Off by default.** `quirks.toolSurface == null` unless a `[providers.*.tool_surface]` table opts in. With it off the request is byte-identical to today (pinned by test + the untouched contract fixtures).
- **Never below status quo.** Cache miss, shape rejection, bad operator config, or a tool whose schema is gone all degrade to today's behavior — never a failed turn. Numeric knobs are clamped at assembly (a `search_limit = 0` typo previously threw).
- **Shape-400 recovery** strips both the `tool_search` tool *and* the invented `tool_search_call`/`tool_search_output` items (plus any dangling `reasoning` item), so the one-shot amend budget retries a clean status-quo-shaped request.
- **Grok untouched.** `forceStrictFalse` (codex parity: `strict:false` on every function tool, matching both reference clients) is deliberately kept **distinct** from grok's pre-existing `emitStrict`; folding them would have silently changed grok's live wire bytes.

## Verification

- `./gradlew check` green; **31/31** ast-grep walls; `config-guard` pass.
- `ToolSurfaceRequestTest` **9/9**, `ToolSurfaceTest` **15/15**, 0 skipped.
- Explicit tests: `input[0]` byte-identical between turn N and turn N+1 (N+1 adds a `tool_use` + `tool_result` of a deferred tool); repeated builds byte-identical; declaration positioned immediately before its `function_call`; deferral-off byte-identical; a transcript `tool_use` naming an absent tool neither crashes nor emits an undeclared call.

## CI

Green. (An earlier revision of this description claimed a pre-existing `webui dist` failure; that was
wrong — it came from a local run with a drifted `node_modules` (react 19.2.7 against a 19.2.8 lockfile).
With a lockfile-synced install the committed bundle reproduces byte-identically, and CI reports
`gate: success`. Retracted in the thread below — notably, "fixing" it would have committed a react
downgrade.)

## Not in this PR

- Live measurement of the realized token saving. `TOOLS_EAGER` / `TOOLS_DEFERRED` are instrumented in the perf JSONL; correlating them against `cached_tokens` on a live head is the empirical confirmation and should happen before enabling this by default anywhere.
- Grok: **do not apply** — no analogue on the xAI chat-completions wire.
- Kimi: **needs its own investigation** — the Anthropic wire has a plausible `tool_search_tool` analogue, worth a cheap live probe.

### DIFF
--- a/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
+++ b/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
@@ -76,9 +76,17 @@ public abstract class ResponsesProvider(
                         null
                     }
                 },
+                // The provider's capability latch, read at build time: false = a shape-400 already
+                // closed it this daemon lifetime; build the full status-quo request instead.
+                toolSurfaceOpen = toolSurfaceLatch.open,
             ),
         )
-        return BuiltTurn(built.req, built.meta, perTurnHeaders(sessionId) + liteHeaders(built.meta))
+        return BuiltTurn(
+            built.req,
+            built.meta,
+            perTurnHeaders(sessionId) + liteHeaders(built.meta),
+            toolSearch = built.toolSearch,
+        )
     }
 
     /** codex-rs sends this marker header for responses-lite (5.6-family) turns; compact turns keep
@@ -137,13 +145,21 @@ public abstract class ResponsesProvider(
     // cache per provider instance; capture and lookup wire in via buildTurn/streamTranslator.
     private val reasoningCache: ReasoningCache = ReasoningCache()
 
+    // The tool-surface capability latch (§1.3/§5.3): one AtomicBoolean per provider instance,
+    // moving in exactly one direction — toward status quo. Read at build time (buildTurn), closed
+    // by amendBodyOnFailure on a shape-400.
+    private val toolSurfaceLatch = ToolSurfaceLatch()
+
     /** RC-4: a 400 rejecting stale encrypted reasoning strips the injected items and retries
-     *  once (NEVER-BELOW-STATUS-QUO law); every other failure keeps the plain retry plan.
+     *  once (NEVER-BELOW-STATUS-QUO law); every other failure keeps the plain retry plan. A 400
+     *  rejecting the tool-surface shape strips the tool_search entry, retries once, and closes the
+     *  latch so every LATER turn on this provider instance builds the full status-quo request.
      *  Keyed off the SAME classifier as the retry plan's GIVE_UP (review 2026-07-24: a narrower
      *  literal match here let any upstream wording drift skip the recovery entirely). */
-    final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? {
-        if (!UpstreamClient.isEncryptedContentError(status, responseText)) return null
-        return stripStaleReasoning(bodyJson, reasoningCache)
+    final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? = when {
+        UpstreamClient.isEncryptedContentError(status, responseText) -> stripStaleReasoning(bodyJson, reasoningCache)
+        isToolSurfaceRejection(status, responseText) -> dropToolSearchTool(bodyJson)?.also { toolSurfaceLatch.close() }
+        else -> null
     }
 
     // The controller is stateless — one cached instance serves every turn (a per-call
--- a/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
+++ b/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
@@ -155,11 +155,32 @@ public abstract class ResponsesProvider(
      *  rejecting the tool-surface shape strips the tool_search entry, retries once, and closes the
      *  latch so every LATER turn on this provider instance builds the full status-quo request.
      *  Keyed off the SAME classifier as the retry plan's GIVE_UP (review 2026-07-24: a narrower
-     *  literal match here let any upstream wording drift skip the recovery entirely). */
+     *  literal match here let any upstream wording drift skip the recovery entirely).
+     *  Honesty gap (review 2026-07-25, [ToolSurfaceLatch]'s KDoc has the full account): the amend
+     *  return value here is eager-only for THIS turn — this function only ever sees
+     *  (status, responseText, bodyJson), never the [ToolPartition] that would let it re-attach the
+     *  deferred tools' schemas, so the recovery turn runs one turn below full status quo before
+     *  the latch restores every later turn. [logToolSurfaceLatchClosed] makes that one-time degrade
+     *  observable instead of silent. */
     final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? = when {
         UpstreamClient.isEncryptedContentError(status, responseText) -> stripStaleReasoning(bodyJson, reasoningCache)
-        isToolSurfaceRejection(status, responseText) -> dropToolSearchTool(bodyJson)?.also { toolSurfaceLatch.close() }
+        isToolSurfaceRejection(status, responseText) -> dropToolSearchTool(bodyJson)?.also {
+            if (toolSurfaceLatch.close()) logToolSurfaceLatchClosed()
+        }
         else -> null
+    }
+
+    /** The latch's one observable signal — stderr, the same `[tag] message` idiom
+     *  ApiKeyAuthProvider.kt uses for standalone diagnostics (no injected logger reaches this
+     *  module; see [ToolSurfaceLatch]). Guarded by [ToolSurfaceLatch.close]'s CAS return, so this
+     *  fires EXACTLY ONCE per provider instance — never once per turn, since every turn after the
+     *  close reads the latch already-closed and never re-enters this branch. */
+    private fun logToolSurfaceLatchClosed() {
+        System.err.println(
+            "[${quirks.providerTag}] tool-surface latch closed: backend rejected the tool_search " +
+                "shape; this turn recovered eager-only (one turn below status quo), every later turn " +
+                "on this provider instance builds the full eager surface.",
+        )
     }
 
     // The controller is stateless — one cached instance serves every turn (a per-call
--- a/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
+++ b/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
@@ -45,6 +45,13 @@ import splice.core.wire.ToolDefinition
 import splice.core.wire.ToolUseBlock
 import java.util.concurrent.atomic.AtomicBoolean
 
+/** Per-provider deferred-tool-surface policy — governs whether/how a request's `tools` gets
+ *  PARTITIONed into an eager set (declared normally in `additional_tools`) and a deferred set
+ *  (answered on demand via `tool_search`; see this file's header). Overlaid from the operator-
+ *  facing TOML `[providers.*.quirks.tool_surface]` table via [withToolSurfaceToml] into
+ *  [ResponsesQuirks.toolSurface]. An ABSENT policy (`toolSurface == null`) means the feature is
+ *  OFF — every request is byte-identical to pre-feature status quo (all tools eager, no
+ *  tool_search object); there is no "configured but inert" state between off and active. */
 public data class ToolDeferralPolicy(
     val deferPrefixes: List<String> = listOf(MCP_PREFIX),
     /** Names FORCED deferred regardless of prefix — the Agent/Task availability brake. */
@@ -53,7 +60,17 @@ public data class ToolDeferralPolicy(
     val eager: Set<String> = emptySet(),
     /** Below this many deferrable tools the split buys nothing and costs a round. */
     val minDeferred: Int = DEFAULT_MIN_DEFERRED,
+    /** Max tools returned by one `tool_search` answer (clamped 1..this —
+     *  [ResponsesToolSearchController.clampedLimit]). Trades answer size against round count: a
+     *  HIGHER limit answers a broad query in fewer rounds at the cost of a bigger
+     *  tool_search_output payload every round; a LOWER limit keeps each round's answer small but
+     *  pushes more queries toward [searchRounds]'s exhaustive fallback. */
     val searchLimit: Int = DEFAULT_SEARCH_LIMIT,
+    /** Permitted `tool_search` rounds before the FINAL round answers with the ENTIRE deferred set
+     *  regardless of query — the loop-can't-wedge law (ResponsesToolSearch.kt header): capability
+     *  at the cap is exactly today's full surface. Trades round budget against when narrowing
+     *  gives up: MORE rounds lets ranked, limit-sized answers keep narrowing longer before the
+     *  exhaustive fallback; FEWER rounds reaches the larger, unranked exhaustive answer sooner. */
     val searchRounds: Int = DEFAULT_SEARCH_ROUNDS,
 )
 
@@ -66,13 +83,29 @@ internal data class ToolPartition(val eager: List<ToolDefinition>, val deferred:
 /** The capability latch — the ENTIRE persistent state of this design: one bit per provider
  *  instance, moving in exactly ONE direction (toward status quo). Closed by a backend rejection
  *  of the tool-surface shape; a daemon restart re-opens it, so a transient backend change costs
- *  one re-probe per lifetime, never a permanent silent downgrade. */
+ *  one re-probe per lifetime, never a permanent silent downgrade.
+ *
+ *  Honesty gap, accepted (review 2026-07-25, options 1+2 over 3): the turn that TRIGGERS the
+ *  close is itself served on a degraded surface, not full status quo. [ResponsesProvider.
+ *  amendBodyOnFailure]'s signature is `(status, responseText, bodyJson)` — the already-serialized
+ *  wire body, never the [ToolPartition] or [ToolDeferralPolicy] that produced it — so the amend
+ *  path can strip the invented tool_search shape but cannot reconstruct and re-attach the
+ *  deferred tools' full schemas (they never rode the request by design; this file's header). That
+ *  ONE recovery turn therefore completes eager-only: one turn below full status quo. Every LATER
+ *  turn on this provider instance reads [open] == false and builds the full eager set from the
+ *  start via [partitionTools]'s `!opts.toolSurfaceOpen -> allEager` branch — true status quo.
+ *  Cost: one degraded-but-successful turn per provider instance per daemon lifetime. Restoring
+ *  the full surface on the amend path itself (option 3) was rejected: it would touch the same
+ *  shared amend seam RC-4's stale-reasoning recovery uses. [close] returns whether THIS call
+ *  performed the actual transition so the caller can fire its one observable signal exactly once
+ *  per instance, never once per amend attempt. */
 internal class ToolSurfaceLatch {
     private val openFlag = AtomicBoolean(true)
     val open: Boolean get() = openFlag.get()
-    fun close() {
-        openFlag.set(false)
-    }
+
+    /** CAS, not a plain set: only the open->closed transition returns true, so a caller racing
+     *  two amend attempts (or any future re-entry) fires its visibility signal exactly once. */
+    fun close(): Boolean = openFlag.compareAndSet(true, false)
 }
 
 /** Overlay the head's TOML `[providers.*.quirks.tool_surface]` table — a DIRECT set, not the
@@ -153,7 +186,11 @@ internal fun functionToolObject(t: ToolDefinition, emitStrict: Boolean, forceStr
         put(FIELD_DESCRIPTION, t.description ?: "")
         put(FIELD_PARAMETERS, t.inputSchema ?: emptyObjectSchema())
         when {
-            forceStrictFalse -> put(FIELD_STRICT, t.strict == true)
+            // forceStrictFalse is a HARD SET (codex-rs parity, responses_api.rs:29-32): every
+            // function tool gets strict:false regardless of the tool's OWN value — passing
+            // t.strict through here (review 2026-07-25) would send strict:true for a tool that
+            // arrives with strict==true, breaking the exact parity this quirk exists for.
+            forceStrictFalse -> put(FIELD_STRICT, false)
             emitStrict && t.strict == true -> put(FIELD_STRICT, true)
             else -> Unit
         }
@@ -172,7 +209,10 @@ internal fun deferredToolObject(t: ToolDefinition, emitStrict: Boolean, forceStr
         put(FIELD_DEFER_LOADING, true)
         put(FIELD_PARAMETERS, t.inputSchema ?: emptyObjectSchema())
         when {
-            forceStrictFalse -> put(FIELD_STRICT, t.strict == true)
+            // Same hard-set law as functionToolObject's identical branch just above (review
+            // 2026-07-25) — a deferred tool answered through tool_search gets forced strict:false
+            // too, never a pass-through of its own strict==true.
+            forceStrictFalse -> put(FIELD_STRICT, false)
             emitStrict && t.strict == true -> put(FIELD_STRICT, true)
             else -> Unit
         }
--- a/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
+++ b/gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
@@ -1,0 +1,264 @@
+// NEW: the deferred tool surface for responses-lite (gpt-5.6 family). splice forwards Claude Code's
+// FULL tool list — measured 64-87 function tools per request, median upstream body 399KB, p90 1.1MB —
+// while OpenAI's own client serves these models a collapsed surface (models.json gives gpt-5.6-sol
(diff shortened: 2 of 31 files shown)


gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
```diff
@@ -0,0 +1,264 @@
...
+ *  into `emitStrict` silently changed grok's live wire bytes too — a head this feature must not
+ *  touch. forceStrictFalse defaults false and only CodexProvider sets it. */
+internal fun functionToolObject(t: ToolDefinition, emitStrict: Boolean, forceStrictFalse: Boolean): JsonObject =
+    buildJsonObject {
+        put(FIELD_TYPE, TYPE_FUNCTION)
+        put(FIELD_NAME, t.name)
+        put(FIELD_DESCRIPTION, t.description ?: "")
+        put(FIELD_PARAMETERS, t.inputSchema ?: emptyObjectSchema())
+        when {
+            forceStrictFalse -> put(FIELD_STRICT, t.strict == true)
```
marcospaulo: **[suggestion]** The forceStrictFalse / emitStrict split is a clean way to keep the codex-only quirk from leaking into grok's wire bytes, and for every tool Claude Code sends today (strict is always null, as this file's own KDoc notes) the emitted bytes are exactly the documented strict:false.

One drift between the documented contract and the code: the KDoc on the quirk (ResponsesRequestBuilder.kt:86-89) and on functionToolObject (ToolSurface.kt:142-148) promise that codex hard-sets strict:false on EVERY function tool regardless of the tool's own value (codex-rs responses_api.rs:29-32), but the branch passes the tool's own value through:

```kotlin
// ToolSurface.kt:156 (functionToolObject) and ToolSurface.kt:175 (deferredToolObject)
forceStrictFalse -> put(FIELD_STRICT, t.strict == true)
```

This fails when a ToolDefinition arrives with strict == true under the codex profile: the upstream request carries strict:true where codex-rs would send strict:false, a parity break in the exact quirk that exists for parity. It is latent today (no client sends strict), and the contract golden `responses codex profile pins strict false on every function tool` (ResponsesContractTest.kt:63) stays green only because its fixture tool has no strict field, so the pin cannot catch the regression.

Aligning the code with the documented contract is one line in each builder:

```kotlin
forceStrictFalse -> put(FIELD_STRICT, false)
```

Optionally, add a fixture tool with `"strict":true` to the codex-profile contract test so the pin covers the case it promises.

One uncertainty worth naming: I have not verified whether any current or planned client path can actually deliver strict:true beyond the KDoc's "always null" claim. If that invariant is enforced at the wire boundary, this is belt-and-braces hardening; if it is merely incidental, the first strict:true tool diverges from codex-rs.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
```diff
@@ -137,13 +145,21 @@ public abstract class ResponsesProvider(
...
+     *  rejecting the tool-surface shape strips the tool_search entry, retries once, and closes the
+     *  latch so every LATER turn on this provider instance builds the full status-quo request.
      *  Keyed off the SAME classifier as the retry plan's GIVE_UP (review 2026-07-24: a narrower
      *  literal match here let any upstream wording drift skip the recovery entirely). */
-    final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? {
-        if (!UpstreamClient.isEncryptedContentError(status, responseText)) return null
-        return stripStaleReasoning(bodyJson, reasoningCache)
+    final override fun amendBodyOnFailure(status: Int, responseText: String, bodyJson: String): String? = when {
+        UpstreamClient.isEncryptedContentError(status, responseText) -> stripStaleReasoning(bodyJson, reasoningCache)
+        isToolSurfaceRejection(status, responseText) -> dropToolSearchTool(bodyJson)?.also { toolSurfaceLatch.close() }
```
marcospaulo: **[suggestion]** The recovery design here is sound: stripping BOTH the tool_search tool and the synthesized call/output items (plus their dangling reasoning) is exactly what makes a rejection of the invented shape recoverable at all, and the one-directional latch keeps the downgrade transient. This comment is about the one-turn surface gap the recovery cannot close, not about the mechanism, which I verified self-terminates cleanly (the amended retry carries no tool_search, so the round outcome has toolSearches empty and the stopSearching isEmpty clause ends the search loop in the same turn).

The gap: when the shape-400 fires, the amended body is rebuilt as

```kotlin
val next = base.copy(input = stripToolSearchItems(toolStripped)) // ToolSurfaceRecovery.kt:40-43
```

and handed back from amendBodyOnFailure (ResponsesProvider.kt:161). What that body cannot contain is the deferred tool schemas. They were removed at build time and never rode the request (by design: deferred tools are absent from the request entirely), and the amend signature (status, responseText, bodyJson) cannot reach the turn's ToolSearchController, the only place those specs still live. So the turn that eats the rejection completes on the EAGER-ONLY surface: the model loses the deferred tools AND the means to discover them (tool_search is stripped too), where full status quo for the same request would have offered the entire list. That is one turn silently below the PR's never-below-status-quo law.

The blast radius is genuinely small, which is why this is a suggestion and not a concern: toolSurfaceLatch.close() means every later turn on this provider instance builds allEager (ToolSurface.kt:93), which IS the full status quo, so the cost is one degraded-but-successful turn per provider instance per daemon lifetime. Falsifiably: drive a responses-lite turn with deferral active against a backend that rejects the shape (status 400-422 mentioning tool_search or defer_loading), and the retried body carries neither the deferred schemas nor tool_search while the turn still reports success.

Three directions, cheapest first:

1. Document the one-turn gap where a reader will look for it (the ToolSurfaceLatch KDoc or amendBodyOnFailure), so "one turn below status quo per daemon lifetime" is a recorded trade-off rather than a discovered one.

2. Make the event visible: a perf counter or log line at the latch close fits the existing expected-delta idiom (TurnMeta.toolsEager/toolsDeferred already stamp the perf JSONL so a false landing is greppable). The amend path does not currently receive perf, so the natural seam is the latch itself, or the provider reading latch state at turn end and stamping the counter there.

3. If the gap is worth closing rather than documenting, the turn's controller already holds every deferred ToolDefinition, so the recovery could rebuild the full eager surface instead of only deleting tool_search. Sketch:

```kotlin
// TurnDriver.postRound wiring: drive.toolSearch is in scope at this call site
amendBodyOnFailure = { status, text, bodyJson ->
    if (isToolSurfaceRejection(status, text)) {
        drive.toolSearch?.restoreFullSurface(bodyJson) // re-emit deferred specs as eager
            ?: provider.amendBodyOnFailure(status, text, bodyJson)
    } else {
        provider.amendBodyOnFailure(status, text, bodyJson)
    }
}
```

(restoreFullSurface would do today's strip of the synthesized items, then overwrite the tools array with eager + deferred, keeping input[0]'s position so the cache prefix moves only as much as a recovery inherently must.)

My read: options 1 and 2 close the honesty gap at near-zero risk; option 3 is only worth its complexity if a degraded turn is ever observed to matter in practice.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesToolSearchTest.kt
```diff
@@ -0,0 +1,231 @@
...
+        val tools = next["input"]!!.jsonArray.last().jsonObject["tools"]!!.jsonArray
+        assertEquals(DEFERRED.size, tools.size)
+    }
+
+    @Test
+    fun `the model's limit is clamped into 1 point policy searchLimit`() {
+        val c = call("ts_1", "tool", limit = 999)
+        val next = CONTROLLER.continuationForSearch(ToolSearchRound(priorRequest, success(listOf(c)), 0))!!
+        val tools = next["input"]!!.jsonArray.last().jsonObject["tools"]!!.jsonArray
+        assertEquals(POLICY.searchLimit, tools.size)
```
marcospaulo, line 181: **[suggestion]** The clamp test pins the upper bound well: limit = 999 clamps down to POLICY.searchLimit, matching the implementation's coerceIn(1, policy.searchLimit) at ResponsesToolSearch.kt:150.

The test name claims the whole range, "clamped into 1..policy.searchLimit", but only the upper side is exercised (ResponsesToolSearchTest.kt:177-182). The lower bound has a real failure mode behind it: if the clamp ever lost its floor (say a refactor to coerceAtMost), a model passing limit = 0 would get an empty tools[] answer for that round, the exact starvation the blank-query branch deliberately avoids, and the suite would stay green.

A second fixture closes it, mirroring the existing test's shape:

```kotlin
@Test
fun `a limit of zero is clamped up to one tool`() {
    val c = call("ts_1", "tool", limit = 0)
    val next = CONTROLLER.continuationForSearch(ToolSearchRound(priorRequest, success(listOf(c)), 0))!!
    val tools = next["input"]!!.jsonArray.last().jsonObject["tools"]!!.jsonArray
    assertEquals(1, tools.size)
}
```

gateway/dialect-openai-responses/src/test/kotlin/ResponsesToolSearchTest.kt
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Test
+    fun `a duplicate call_id in one round is answered exactly once`() {
+        val calls = listOf(call("ts_1", "tool_0"), call("ts_1", "tool_1"))
+        val next = CONTROLLER.continuationForSearch(ToolSearchRound(priorRequest, success(calls), 0))!!
+        val tail = appendedTail(next)
+        val outputs = tail.map { it.jsonObject }.filter { it["type"]?.jsonPrimitive?.content == "tool_search_output" }
+        assertEquals(1, outputs.size)
+    }
```
marcospaulo, line 210: **[suggestion]** This test pins the most important observable of the dedup rule (a duplicated call_id produces exactly one tool_search_output), and the first-wins contract is documented where it lives, in answeredOnce's KDoc at ResponsesToolSearch.kt:136.

The fixture gives the two duplicates different queries ("tool_0" vs "tool_1"), but the assertion only counts the outputs (ResponsesToolSearchTest.kt:185-191). That leaves WHICH duplicate gets answered unpinned: if answeredOnce ever regressed from first-wins to last-wins, the wire would carry the tool set for "tool_1" instead of "tool_0" and this test would stay green.

Since the two queries are already distinguishable, one more assertion locks the contract. Sketch (exact expected names depend on the DEFERRED fixture):

```kotlin
val tools = outputs.single().jsonObject["tools"]!!.jsonArray
// first-wins: the answered set is the one the FIRST call's query ("tool_0") ranks
assertEquals(DEFERRED.first { it.name.contains("tool_0") }.name,
    tools.first().jsonObject["name"]?.jsonPrimitive?.content)
```

Any assertion that inspects the returned tools against the first query works here; the point is only that the count check alone cannot see a first/last swap.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
```diff
@@ -0,0 +1,264 @@
...
+import kotlinx.serialization.json.add
+import kotlinx.serialization.json.buildJsonArray
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
+import splice.core.wire.AnthropicRequest
+import splice.core.wire.ToolDefinition
+import splice.core.wire.ToolUseBlock
+import java.util.concurrent.atomic.AtomicBoolean
+
+public data class ToolDeferralPolicy(
```
marcospaulo, line 55: **[nit]** The documentation culture in this file is strong: every neighboring public declaration carries dense KDoc, and four of this policy's six parameters already document their rationale.

ToolDeferralPolicy itself (ToolSurface.kt:48) is the exception: no class-level KDoc, and searchLimit / searchRounds (ToolSurface.kt:56-57) are the only undocumented parameters. Since the module is under explicitApi and this type rides ResponsesQuirks.toolSurface out to the operator-facing TOML overlay, the class is part of the deliberate public surface, and a reader currently gets the knobs without the contract (what the policy governs, that an absent policy means off, and what the two search bounds trade off: answer size per round vs rounds before the exhaustive final answer).

A short header plus two param docs closes it, something like:

```kotlin
/** The deferred tool surface policy: which tools leave the eager payload, when the split buys
 *  nothing (minDeferred), and how the gateway answers tool_search (searchLimit per round,
 *  searchRounds before the final round answers with the entire deferred set). */
```

Pure preference at this severity; nothing here is wrong, just slightly less self-describing than the surface around it.

marcospaulo (commented):
**Review of:** deferred tool surface with cache-stable gateway-answered `tool_search` on openai-responses heads (5 inline comments: 4 medium, 1 nit; 39 raw findings dissolved by evidence gates, reasons recorded).

**What was verified clean:**

- **Security lens** (12 files): synthesized wire items are serializer-built (no JSON injection), search query matching is plain `contains` (no regex injection), config knob normalizes to off/auto (unknown values never silently arm the feature), no credential or request-body logging added, no disk/state writes.
- **Mechanical scan**: 38 ast-grep rules ran (bundled + this repo's own packs); 30 hits, all `kt-no-notnull-assert` in test files, all dissolved as false positives (fixture values guarded by preceding asserts). Signal for the rules owners: consider scoping that rule to main sources.
- **Flow-map seam check** (abort/cancellation/amend-budget state across 14 files): the shape-400 recovery was suspected of looping the current turn through repeated 400s. Verified self-terminating: the amended body is local to the retry, the recovered outcome carries no `tool_search_call`s, and the `toolSearches.isEmpty()` stop clause ends the search loop in the same round. One related finding survived (see ToolSurfaceRecovery comment).
- **API/ABI lens**: 6 binary-compatibility candidates dissolved after verifying no module is published or ABI-validated (only `explicitApi()` source-level discipline in `splice.module-law.gradle.kts`); the app compiles atomically.
- **The known-red `webui dist` CI**: consistent with the PR's claim of pre-existing (this branch changes 0 files under `webui/`).

**Test coverage observations (not blocking, offered for follow-up):**

- `deferredToolObject` under `forceStrictFalse=true` (and `emitStrict && strict=true`) is not exercised by any test; related to the F044 finding below.
- The latch-close path (`amendBodyOnFailure -> dropToolSearchTool -> latch.close`) has no direct test.
- `dropToolSearchTool` with multiple contiguous reasoning items or multiple search-call/output pairs is untested.
- The 30 `!!` occurrences in the new tests are stylistic; in each case the nullable is guarded by a preceding assertion.

**Antibody drafts:** none (no new mechanically-expressible defect class survived the gates).

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
marcospaulo: Fixed in 99f9adf — you were right that this is a real parity break, not just doc drift.

Both builders now emit the constant:

```kotlin
// ToolSurface.kt:193 (functionToolObject) and :215 (deferredToolObject)
forceStrictFalse -> put(FIELD_STRICT, false)
```

Your point about the pin being unable to catch it was the sharper half of the finding — the golden stayed green only because its fixture tool had no `strict` field. The codex-profile fixture now carries a second tool with `"strict": true`, so the test covers the case it promises (both tools emit `"strict": false`).

On your named uncertainty: I did not find any client path that can deliver `strict:true` today either — `ToolDefinition.strict` is populated only from the Anthropic wire and Claude Code never sets it. So this is belt-and-braces rather than a live break, but the code now matches its documented contract instead of depending on that incidental invariant.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
marcospaulo: Taking options 1 and 2, deliberately not 3 — your reasoning that option 3 touches the shared amend seam (RC-4 stale-reasoning recovery rides the same path) is exactly why. One degraded-but-successful turn per provider instance per daemon lifetime does not justify restructuring that seam.

In 99f9adf:

1. **Documented** at both sites — the `ToolSurfaceLatch` KDoc and `amendBodyOnFailure` now state that the recovery turn runs eager-only (one turn below status quo), that every later turn is full status quo via the latch, and *why* the deferred specs are unreachable: `amendBodyOnFailure(status, responseText, bodyJson)` never sees the turn ToolPartition.

2. **Made visible** — the latch close was silent, which was the real defect. `close()` is now a CAS returning the open→closed transition, and the signal is emitted exactly once per provider instance (not once per turn). `CodexProviderTest` pins this: it captures stderr across *two* amend attempts and asserts the signal fires once.

That test also covers the rest of your falsifiable scenario end to end: a shape-rejection through the real provider strips the tool_search tool plus the synthesized call/output/reasoning items, and a subsequent `buildTurn` on the same instance goes 10-deferred/1-eager → 0-deferred/11-eager, pinning the status-quo restoration.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesToolSearchTest.kt
marcospaulo: Fixed in 99f9adf — added `a limit of zero is clamped up to one tool`, so the test name now matches what is exercised on both sides.

Agreed on the failure mode: a refactor to `coerceAtMost` would answer `limit = 0` with an empty `tools[]` — the same starvation the blank-query branch deliberately avoids — and the suite would have stayed green.

gateway/dialect-openai-responses/src/test/kotlin/ResponsesToolSearchTest.kt
marcospaulo: Fixed in 99f9adf — the dedup test now asserts the answered set is the one the FIRST call query ranks (`mcp__exa__tool_0`), so a first-wins → last-wins regression fails instead of passing.

Checked the substring risk you implied by using distinguishable queries: `"tool_0"` matches only `mcp__exa__tool_0` in the DEFERRED fixture (no collision with `tool_10`/`tool_11`), so the assertion is not brittle.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ToolSurface.kt
marcospaulo: Fixed in 99f9adf — added class-level KDoc plus param docs for `searchLimit` and `searchRounds`, covering what the policy governs, that an absent policy means OFF, and the answer-size-per-round vs rounds-before-the-exhaustive-answer trade-off.

Fair call that this is public surface rather than an internal detail: it rides `ResponsesQuirks.toolSurface` out to the operator-facing TOML overlay under `explicitApi()`, so the knobs were reaching operators without the contract.

marcospaulo:
All 9 items addressed in 99f9adf — the 5 inline comments plus the 4 test-coverage observations from the review body. Gate green: `gradlew check` + 31/31 ast-grep walls + config-guard.

**The one real bug:** `forceStrictFalse` passed the tool own `strict` through in both builders while its KDoc promised a hard-set — and the golden that should have caught it could not, because its fixture tool had no `strict` field. Both fixed. Thank you for pushing past the code to the test that was supposed to be guarding it.

**Review-body items (marked non-blocking, done anyway):**

- **`deferredToolObject` strict handling untested** — now covered in all four combinations. Deliberately includes a case pinning that grok pre-existing `emitStrict` pass-through is **unchanged**, since that separation is the whole reason the two quirks are distinct.
- **Latch-close path untested** — `CodexProviderTest` drives a real `CodexProvider` through `amendBodyOnFailure`: asserts the strip, asserts the subsequent build goes 10-deferred/1-eager → 0-deferred/11-eager (the status-quo restoration), and captures stderr across two amend attempts to assert the new signal fires exactly once. The instrumentation is pinned, not merely present.
- **`dropToolSearchTool` multi-item cases untested** — added for a contiguous reasoning run and for multiple call/output pairs. I asked for actual behavior to be pinned rather than the code bent to match my phrasing; the backward walk turned out to already strip the entire contiguous run and the forward scan already handles multiple pairs, so these are pure coverage, no behavior change.
- **The 30 `!!` in tests** — left alone, agreeing with your false-positive verdict (each guarded by a preceding assertion or a fixed literal).

**One item consciously not actioned:** your signal that `kt-no-notnull-assert` be scoped to main sources. Rules are grant-gated in this repo, and narrowing one inside a feature PR is exactly the kind of change that should not ride along unnoticed — surfaced to the operator as its own decision.

**Also noting your verification work**, since it shaped what I did *not* do: the flow-map finding that the shape-400 recovery self-terminates (amended body carries no `tool_search_call`s → `toolSearches.isEmpty()` ends the loop in-round) is what made documenting-and-instrumenting the right call over restructuring the shared amend seam. And the confirmation that `webui dist` red is pre-existing matches my own trace — the committed bundle was built ~14h before the toolchain bump, deterministic, 6 bytes apart. That fix belongs on `main`; happy to land it separately whenever you want CI green.

Still open from the PR description, unchanged by this round: the token saving is proven structurally (schemas absent from the wire, `input[0]` byte-stable across turns) but **not yet measured live** — `TOOLS_EAGER`/`TOOLS_DEFERRED` are instrumented for exactly that, and it should happen before this is enabled by default anywhere.

marcospaulo:
**Correction — the "known-red CI" claim in my PR description is wrong. Retracting it.**

The reviewer chased this properly and the premise does not survive:

- **CI has been green.** `gate: success` on `99f9adf`; `lint`/CodeQL/dependency-review green on both heads. There is no failing webui-dist run.
- **The committed bundle was never stale.** With a lockfile-synced install, `vite build` reproduces `webui/dist/index.html` **byte-identically**. I re-verified independently just now: 0 differing bytes.
- **The 6-byte diff was my environment.** My `node_modules` had react 19.2.7 against a 19.2.8 lockfile. Building with a drifted install produces exactly that diff and mimics "deterministic build, stale committed bundle."

**So: do not cut a rebuild commit — not here, not on `main`.** Committing that bundle would have pushed a **react downgrade** into `dist` disguised as a hygiene fix.

Where my reasoning failed, since it is the useful part: two builds agreeing with each other proved the build was deterministic, not that the *committed* artifact was stale — both of my builds shared the same drifted `node_modules`. I compared build-against-build and concluded commit-is-stale. The tell was in my own output: one manual `cmp` passed and the next gate run failed. I noted the inconsistency and moved on instead of chasing it, and it was pointing straight at my install.

`git diff main...HEAD -- webui/` being empty was the only half of that claim I actually verified; I took the red half on report from a local gate run and did not check CI. My mistake, and it is exactly the failure mode this PR review process is meant to catch.

The **`## Known-red CI`** section of the PR description is now inaccurate and I am editing it out.

Unchanged and still the one real gap before flipping any default: **live measurement of the token saving** — `TOOLS_EAGER`/`TOOLS_DEFERRED` correlated against `cached_tokens` on a live head.

marcospaulo (commented):
Verification pass on 99f9adf (would be an approval, but GitHub blocks self-approval from the author account): both builders hard-set strict:false with the golden now pinning a strict:true input; the latch close is a CAS firing exactly once with the one-turn honesty gap documented at both sites; the clamp and first-wins pins are real; the new strict/recovery test file covers the gaps the review summary called out. CI green on both heads. All review threads resolved. Good to merge — the deferred-surface live measurement (TOOLS_EAGER/TOOLS_DEFERRED vs cached_tokens) remains the gate before enabling this anywhere, correctly separate from merge.

PR #49, opened 2026-07-25
fix(gateway): recover 16 unmerged hardening fixes + two generation-time walls

marcospaulo:
Recovers 16 gate-verified hardening fixes that were **never merged**, and adds two generation-time walls so the same class of loss is caught mechanically next time.

## The incident

Three hardening commits (`be1baee` / `473b7ff` / `1acceec` — 16 fixes across auth, wire, gateway and config, each gate-verified when written) sat on a local branch. `main` advanced independently through the PR #48 squash (`c0a9161`), cut from the same base — so **every hardening line was absent from `main`**, and from any build made from it.

Found by an audit of the working session's own transcript, not by any check in this repo. That is the point of the walls below.

## Live on `main` until this merge

- `TurnDriver`: raw `e.message` interpolated into the **client** error frame; unbounded upstream `failure.message`; nullable `(null)` conn-reset detail
- `PassthroughRequestBuilder`: unguarded `stripCacheControl` recursion → StackOverflow on adversarially nested client JSON
- `ChatStreamTranslator` / `ResponsesStreamTranslator`: unbounded accumulation buffers
- `AsyncFileIo`: silently dropped writes · `ConfigService`: swallowed persist failure + `(mtime,size)` torn read · `CodexAuthProvider`: mtime-only auth cache
- `OAuthLoginFlow`: provider error body echoed to operator stdout

## Merge detail

One real conflict, in `ResponsesStreamTranslator`'s companion: `main` had moved `REASONING_KEY_BASE`/`reasoningKey` to file top-level (a detekt extraction in #48) while the hardening branch added `MAX_DEDUP_SPLIT_CHARS` beside them. Resolved by keeping `main`'s structure and carrying only the WIRE-4 guard across.

## The two walls (concept #924)

Both defects were produced by agents in a single session. Per #924's corollary the bar is *"what holds when agents keep writing it"*, not *"the code is clean now"* — so each is encoded as a rule that runs at **write time** (PreToolUse, against proposed content) and again at the gate, per `sgconfig.yml`'s same-checker-twice doctrine.

- **`kt-force-strict-false-is-literal`** — a `forceStrictFalse` branch must emit the literal `false`. It shipped as `put(FIELD_STRICT, t.strict == true)` in both tool builders: the KDoc promised a codex-parity hard-set, the code passed the tool's own value through. It survived the gate because the contract golden meant to pin it used a fixture tool with no `strict` field, so the assertion could not observe the difference.
- **`kt-tool-partition-no-transcript`** — the deferred-tool partition must be a pure function of `(tools, policy)`. The payload rides `input[0]` inside the prompt-cache prefix, so a transcript-derived split varies it per turn and re-bills ~94k cached tokens per bust — silently turning #48 into a net token *regression*. Three independent adversarial reviews missed it; all verified bytes-when-stable, none turn-to-turn.

Both were validated with `ast-grep test_match_code_rule` before wiring, then proven red-green. Generation-time behavior was proven by probe: a `Write` carrying the shipped defect was **blocked and never landed**.

## Verification

- `gradlew check` green on the merged tree
- ast-grep walls **33 passed / 0 failed** (was 31), scan clean, hook tests 13 OK, `config-guard` PASS
- Fixes confirmed present in the built artifact, not just the source: the jar contains `internal gateway error — retry` and not the interpolated form

## Note for the reviewer

The lesson worth carrying: **an edit protects one branch; a wall protects the generator.** Sixteen verified fixes went missing precisely because they were only edits. The pattern-shaped findings among them (raw detail on the client wire, unguarded recursion over client JSON, unbounded grow-only buffers) deserve walls of their own — `kt-no-raw-detail-on-client-wire` is drafted and hand-verified against all 8 `emitError` call sites (2 hits, both real defects, 6 compliant sites silent), pending an operator grant since `.rules/` is grant-gated.

### DIFF
--- a/.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
+++ b/.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
@@ -30,6 +30,12 @@ rule:
       kind: when_condition
       regex: '^forceStrictFalse$'
 constraints:
+  # FIELD is pinned (review #49): unconstrained, the rule fired on ANY put in the branch — a
+  # future compliant edit that groups another builder call into it (e.g. put(FIELD_DEFER_LOADING,
+  # true)) would be blocked mid-write. A wall that fires on compliant code teaches agents that
+  # wall output is noise, which is the failure #924 is about.
+  FIELD:
+    regex: '^FIELD_STRICT$'
   VAL:
     not:
       regex: '^false$'
--- a/.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
+++ b/.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
@@ -1,0 +1,35 @@
+# VENDOR-PARITY WALL (2026-07-25): a `forceStrictFalse` branch must emit the LITERAL false.
+# The quirk exists for ONE reason — codex-rs hard-sets `strict:false` on EVERY function tool
+# regardless of the tool's own value (responses_api.rs:29-32; OpenCode does the same, marked
+# "Codex parity"). It SHIPPED as `put(FIELD_STRICT, t.strict == true)` in BOTH tool builders:
+# the KDoc promised a hard-set, the code passed the tool's own value through. A ToolDefinition
+# with strict=true would have sent strict:true where codex-rs sends false — a parity break in
+# the very quirk that exists for parity.
+#
+# It survived the gate because the contract golden meant to pin it ("pins strict false on every
+# function tool") used a fixture tool with NO strict field, so the assertion could not observe
+# the difference. Human review caught it; nothing mechanical did. This is the generation-time
+# half (PreToolUse scans the PROPOSED write) of that invariant; the gate half is the contract
+# fixture, which now carries a strict:true tool.
+id: kt-force-strict-false-is-literal
+language: kotlin
+severity: error
+message: "A forceStrictFalse branch must emit the LITERAL false — passing the tool's own strict value through defeats the codex-parity hard-set the quirk exists for (shipped bug, 2026-07-25)."
+note: >
+  Write `put(FIELD_STRICT, false)`. If a provider genuinely needs the tool's own value passed
+  through, that is the SEPARATE `emitStrict` quirk — do not overload forceStrictFalse, which is
+  defined as an unconditional hard-set (ResponsesQuirks KDoc, ResponsesRequestBuilder.kt).
+files:
+  - gateway/*/src/main/**/*.kt
+rule:
+  pattern: put($FIELD, $VAL)
+  inside:
+    kind: when_entry
+    stopBy: end
+    has:
+      kind: when_condition
+      regex: '^forceStrictFalse$'
+constraints:
+  VAL:
+    not:
+      regex: '^false$'
--- a/gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
+++ b/gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
@@ -79,7 +79,8 @@ public class ChatStreamTranslator(private val ctx: ChatTurnContext) : StreamTran
     override suspend fun driveTurn(upstream: Flow<JsonObject>, sink: WireSink): TurnOutcome {
         try {
             upstream.collect { evt ->
-                if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolIndexById.size)) {
+                val toolCount = maxOf(toolIndexById.size, pendingTools.size)
+                if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount)) {
                     runawayGuard = RUNAWAY_GUARD_MESSAGE
                     return@collect
                 }
--- a/gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
+++ b/gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
@@ -80,7 +80,12 @@ public class ChatStreamTranslator(private val ctx: ChatTurnContext) : StreamTran
         try {
             upstream.collect { evt ->
                 val toolCount = maxOf(toolIndexById.size, pendingTools.size)
-                if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount)) {
+                // pendingTools' per-entry `args` accumulate BEFORE the tool opens (deferred-open:
+                // a backend may stream index+id and large argument deltas but never function.name),
+                // so an entry count alone leaves those chars unbounded — review #49. Normally a
+                // handful of entries; the count cap bounds the sum's worst case.
+                val pendingArgsChars = pendingTools.values.sumOf { it.args.length }
+                if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount, pendingArgsChars)) {
                     runawayGuard = RUNAWAY_GUARD_MESSAGE
                     return@collect
                 }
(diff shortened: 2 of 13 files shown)


.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
```diff
@@ -0,0 +1,35 @@
...
+files:
+  - gateway/*/src/main/**/*.kt
+rule:
+  pattern: put($FIELD, $VAL)
+  inside:
+    kind: when_entry
+    stopBy: end
+    has:
+      kind: when_condition
+      regex: '^forceStrictFalse$'
```
marcospaulo, line 31: **[suggestion]** The wall's defect targeting is well chosen: scoping to a `forceStrictFalse` when-entry plus the `VAL != false` constraint pins exactly the shipped pass-through shape, and the rule tests cover both the defect and the compliant literal.

One precision gap: `$FIELD` is unconstrained, so the rule fires on ANY `put` with a non-`false` value inside the branch, not only the strict field it exists to protect. I verified this by running the rule: a branch like

```kotlin
forceStrictFalse -> {
    put(FIELD_STRICT, false)
    put(FIELD_DEFER_LOADING, true)
}
```

gets flagged on the `FIELD_DEFER_LOADING` line even though the wall's message only governs `FIELD_STRICT`. Since this rule runs at write time and at the gate against every future edit under `gateway/*/src/main`, a compliant future change that groups another builder call into the branch would be blocked mid-write.

Constraining the field metavariable keeps the wall exactly as strict on the defect while silencing the over-match:

```yaml
constraints:
  FIELD:
    regex: '^FIELD_STRICT$'
  VAL:
    not:
      regex: '^false$'
```

I re-ran the rule with this constraint: it still fires on `put(FIELD_STRICT, t.strict == true)` and stays silent on both the compliant literal and the grouped `FIELD_DEFER_LOADING` case above. Worth also adding that grouped-put shape to the rule test's valid list so the precision is pinned.

Scoping note: no current code trips this (both call sites in ToolSurface.kt keep the branch single-statement), so this is a future-write hazard rather than a present violation, which is why it lands as a suggestion.

gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
```diff
@@ -72,7 +78,14 @@ public class ChatStreamTranslator(private val ctx: ChatTurnContext) : StreamTran
 
     override suspend fun driveTurn(upstream: Flow<JsonObject>, sink: WireSink): TurnOutcome {
         try {
-            upstream.collect { evt -> onEvent(evt, sink) }
+            upstream.collect { evt ->
+                val toolCount = maxOf(toolIndexById.size, pendingTools.size)
+                if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount)) {
```
marcospaulo: **[suggestion]** The runaway guard is a good addition: bounding textBuf/thinkingBuf at 20M chars and tool-entry counts at 50K converts a runaway upstream from heap exhaustion into a clean turn failure, and the comment is honest about why these buffers must stay unbounded on the normal path.

One buffer escapes it. Each `pendingTools` entry holds an `args` StringBuilder that accumulates streamed arguments (line 276, `pending.args.append(args)`), and the guard reads only entry counts (`maxOf(toolIndexById.size, pendingTools.size)`), never the accumulated chars. The deferred-open design means a tool call can legitimately stream arguments before its name arrives (comment at lines 68-70); a backend that sends index+id and then large argument deltas but never `function.name` leaves one pending entry growing without bound while `toolCount` stays 1 and both char caps read nominal. The watchdog caps time, not bytes, so a fast upstream can still exhaust the heap inside the time cap. That is the same runaway-upstream scenario the guard's own comment says it covers.

Summing the pending arg lengths into the check closes it:

```kotlin
upstream.collect { evt ->
    val toolCount = maxOf(toolIndexById.size, pendingTools.size)
    val pendingArgsChars = pendingTools.values.sumOf { it.args.length }
    if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount, pendingArgsChars)) {
        runawayGuard = RUNAWAY_GUARD_MESSAGE
        return@collect
    }
    onEvent(evt, sink)
}
```

with `bufferOverCapacity` comparing `pendingArgsChars` against the same MAX_BUFFERED_CHARS (a smaller dedicated cap, or a per-tool bound, also works and localizes the failure to one call).

Two scoping notes to keep this honest. First, once the name arrives the tool opens, the buffered args flush, and later chunks stream straight to the sink unbuffered (line 268), so the unbounded window is specifically the never-named-pending case; I am not claiming a well-behaved stream can hit this. Second, the same count-only blind spot covers `toolBlocks` growth from explicitly-indexed calls, though those entries are small. I also checked the ResponsesStreamTranslator sibling: it opens tool blocks eagerly on output_item.added and streams args directly to the sink in onArgs, so it has no equivalent accumulator and does not share this gap.

marcospaulo (commented):
**Review of:** recovery of 16 unmerged hardening fixes + two generation-time walls. 2 inline comments, both medium; 2 raw findings dissolved by evidence gates, reasons recorded.

**Recovery verified against the original commits, not the description:** all 24 hunks of `be1baee^..1acceec` are present in the head content and unweakened (the 16 logical fixes they form). The `ResponsesStreamTranslator` conflict resolution is coherent: `main`'s top-level `REASONING_KEY_BASE`/`reasoningKey` structure and the hardening `MAX_DEDUP_SPLIT_CHARS` guard both survive.

**Mechanical scan:** 40 ast-grep rules ran (bundled + repo packs + this PR's 2 new walls). One hit: `kt-no-println` at `OAuthLoginFlow.kt:236`, dissolved as a false positive: the repo's own rule explicitly ignores `gateway/app/src/main` (the CLI/entry layer, "a terminal tool legitimately writes to stdout"), and the line is the sanitized fix itself (status-only, provider body no longer echoed). The two new walls were silent on this PR's own Kotlin, as expected.

**Wall validation:** both new rules were checked against their rule-test files. `kt-force-strict-false-is-literal` matches the shipped bug shape and stays silent on the compliant literal form (one over-match finding survived, inline). `kt-tool-partition-no-transcript` fires on the realistic regression shapes and stays silent on the adjacent legitimate `declarationCandidates`; it is a syntactic proxy, like all 33 walls, and the claim that it misses other accessors did not hold (those accessors do not exist on `AnthropicRequest`).

**Checked clean:** client-wire error content in TurnDriver (provider messages bounded, runtime messages no longer reach the client), the OAuthLoginFlow/ConfigService/AsyncFileIo/JsonlSink/CodexAuthProvider recoveries, and the sibling `ResponsesStreamTranslator` runaway guard (streams args to sink, does not share the ChatStreamTranslator gap).

**Antibody drafts:** none.

.rules/kotlin-splice/kt-force-strict-false-is-literal.yml
marcospaulo: Fixed in 9af22d2 — you were right to run it rather than read it.

`$FIELD` is now pinned:

```yaml
constraints:
  FIELD:
    regex: "^FIELD_STRICT$"
  VAL:
    not:
      regex: "^false$"
```

Re-validated: still fires on `put(FIELD_STRICT, t.strict == true)`, silent on the compliant literal, and silent on your grouped `FIELD_DEFER_LOADING` case. I also added that grouped-put shape to the rule-test `valid:` list as you suggested, so the precision is pinned rather than merely fixed.

Worth naming why this one mattered more than its severity suggests: a wall that fires on compliant code is the exact failure #924 warns about — it trains agents to treat wall output as noise and route around it. This PR itself contains a finding of that shape (30/30 false positives from a bundled rule, dissolved wholesale). Shipping a second one in the same PR would have been careless.

gateway/dialect-openai-chat/src/main/kotlin/splice/dialect/chat/ChatStreamTranslator.kt
marcospaulo: Fixed in 9af22d2. This was a real gap in the r1 guard, not just a hardening nit — the guard's own comment claims to cover runaway upstreams, and this path escaped it.

```kotlin
val pendingArgsChars = pendingTools.values.sumOf { it.args.length }
if (bufferOverCapacity(textBuf.length, thinkingBuf.length, toolCount, pendingArgsChars)) {
```

with `pendingArgsLen` compared against the same `MAX_BUFFERED_CHARS`.

Your two scoping notes were both load-bearing and I kept them in the code comment: the unbounded window is specifically the never-named-pending case (once the name arrives the tool opens and later chunks stream unbuffered), and the entry-count cap bounds the sum's worst case, so the per-event `sumOf` stays cheap on the normal path where `pendingTools` holds a handful of entries.

I took the simple shape you proposed over a per-tool bound or a running total: it is the one you validated, and a running total would have required touching `applyToolCall`'s append site — more surface for a guard whose whole job is to be boring.

Also confirming your `ResponsesStreamTranslator` check independently: it opens tool blocks eagerly on `output_item.added` and streams args straight to the sink in `onArgs`, so there is no equivalent accumulator and no change was needed there.

marcospaulo (commented):
Verification pass on 9af22d2 (would be an approval, but GitHub blocks self-approval from the author account): the FIELD_STRICT constraint is exactly the reviewed fix, with the grouped-put shape added to the rule test's valid list so the precision is pinned, and the wall edit went through the prescribed grant path. The runaway guard now sums pendingTools args lengths into bufferOverCapacity under the same 20M cap, closing the never-named-pending window while the count cap bounds the sum's worst case. Both review threads resolved. Good to merge once the gate finishes green.

PR #58, opened 2026-07-26
feat(responses): turn-scoped summary dedup — detailed reasoning on every round, zero duplicates

marcospaulo:
> **Scope note (2026-07-27, addressing the review's scope finding).** This PR no longer matches its
> original description. It was cut for one fix and has since accumulated **19 commits from four
> concurrent sessions sharing this checkout** — +11,454 / −115 across 129 files. The body below now
> describes what is actually here. Nothing was force-pushed or rewritten; the growth is real and
> reviewable, but **it is worth deciding whether this should merge as one PR or be split** before
> hitting the button.

## 1. Turn-scoped summary dedup — the original subject

**Your call 2026-07-26:** keep the detailed reasoning summary on **every** round, continuations
included — and no duplicates. This does both; the #57 strip is reverted.

Continuation rounds (fold / reanchor / tool_search) re-request the detailed summary over
already-summarized reasoning, and the backend re-titles it with the same generic section titles.
Each round built a *fresh* translator with *fresh* dedup state, so an exact repeat passed every
round's per-instance dedup and landed as a duplicate block (measured 89–102 per claudex session).

`SharedSummaryParts` (TurnMeta) hoists the emitted-section set to turn scope. It is now **non-null
with a fresh default**, so no caller constructs one and `copy()` — which every continuation path
uses — preserves the reference. Wall `kt-shared-summary-parts-single-source` makes a second
construction unwritable.

## 2. Per-head knob overrides — `[heads.<key>.overrides]`

Heads share one `ConfigService` in one JVM, so a knob tuned for one upstream governed all of them
(kimi's 40-min `upstreamTimeoutMs` also capped codex). Pinned by `DaemonPerHeadConfigTest` and by
wall `kt-head-scoped-config-must-be-keyed`, which states the invariant relationally — *inside a
head-scoped function the unkeyed view is banned* — so it needs no exception list and covers new
call sites automatically.

## 3. Early-open handoff — `perf(stream)`

The turn opens on the wire at upstream handoff instead of at first content block, recovering ~2840ms
p50 of dead air per codex turn.

## 4. ast-grep walls audit — 33 → 36 walls, 29 rule improvements

Nearly every wall was narrower than its own message claimed: `kt-no-sealed-interface` was defeated by
a leading annotation or a double space; `kt-catch-swallows-cancellation` accepted a *comment*
mentioning `CancellationException` as proof of handling. Full audit in
`dev/research/ast-grep-walls-audit-2026-07-26.md`. Two real bugs found by the new walls
(`ControlServer.kt` per-head config; a false positive in `kt-no-object-command`).

Also adds `/grant` + `/install-walls`: wall grants issued from the prompt line, bounded and
revocable, replacing an environment variable that required restarting the CLI.

## Review findings — all 7 closed

| # | finding | closed by |
|---|---|---|
| 1 | summary-dedup wiring unpinned | `af57b0a` — end-to-end through the real head |
| 2 | tear path frame contract | `61d83de`, corrected in `24eefbf` |
| 3 | `[defaults]` referenced but undefined | `61d83de` + a mechanical knob-name check |
| 4 | `resets_in_seconds` precedence untested | `61d83de` |
| 5 | `SharedSummaryParts` access discipline | `61d83de` |
| 6 | per-head wiring unpinned | `61d83de` — `DaemonPerHeadConfigTest` |
| 7 | `SharedSummaryParts` contract comment | `61d83de` |

**One correction on my own work:** the assertion added for finding #2 was a flake — it demanded
`message_start` be *present* on the torn path, which depends on where the premature EOF surfaces.
It passed 3/3 locally and turned CI red. `24eefbf` pins the invariant (ordering: `message_start`
never *after* the error frame) and leaves presence to the deterministic holdstart test.

**Still open, needs an operator decision:** `kt-no-system-getenv` (+21 live hits) and
`kt-no-println` (+13). Both would turn CI red for sites using the same idiom the rules' own exempted
file uses deliberately — see §2.1/§2.2 of the audit.


gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
```diff
@@ -118,6 +118,7 @@ public abstract class ResponsesProvider(
                 streamIdleMsForMessage = watchdog.streamIdle.inWholeMilliseconds,
                 upstreamTimeoutMsForMessage = watchdog.totalCap.inWholeMilliseconds,
                 dedupeRepeatedSummaryParts = quirks.summaryDelivery != null,
+                summaryPartsShared = meta.summaryParts,
```
marcospaulo, line 121: **[concern]** The mechanism is well pinned: `SummaryTurnDedupTest` proves repeats die turn-scoped when the state is shared, and pins the pre-fix bug against two isolated translators. The one thing nothing pins is the wiring that makes the state shared in production.

`ResponsesProvider.kt:121`:
```kotlin
summaryPartsShared = meta.summaryParts,
```

The whole fix rides on a single invariant: every continuation round's translator receives the same `SharedSummaryParts` instance. Today it holds (the builder creates the instance once in `buildTurn`, `TurnDriver` stores `built.meta` on the drive and passes `drive.meta` into `streamTranslator` on every round), but no test exercises that path: `SummaryTurnDedupTest` hand-constructs the shared instance and passes it to two translators directly, and `ResponsesFoldTest` only asserts the summary request survives `continuationRequest`. Falsifiably: if a future change rebuilds the meta per round, or slips an explicit `copy(summaryParts = SharedSummaryParts())` into the fold loop, the 2026-07-26 mirror duplication returns and every test in this PR stays green.

A test that drives the wiring rather than the mechanism would close the gap. Sketch, at whichever level has the cheapest harness (provider or driver):

```kotlin
@Test
fun `continuation rounds share one summaryParts instance`() {
    val metas = mutableListOf<TurnMeta>()
    val provider = recordingProvider { meta -> metas += meta; fakeRoundOutcome() }
    driveTwoRoundContinuation(provider) // fold or re-anchor path
    assertEquals(2, metas.size)
    assertSame(metas[0].summaryParts, metas[1].summaryParts)
}
```

Reference identity is the point of the assertion: `SharedSummaryParts` is never compared by value, and a plain `copy()` keeps the reference, so only a rebuild or an explicit replacement breaks it, which is exactly the kind of change that would wave through review today. Pinning the identity is the requirement; where the test lives is preference, whichever harness is cheaper.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesStreamTranslator.kt
```diff
@@ -100,10 +105,13 @@ private const val RECAP_DONE = -1
 // (2026-07-20); a per-ITEM set alone under-suppressed the cross-item recap (the duplication
 // staircase). Splitting the two jobs keeps the coincidence (per-item, non-leading) while killing
 // the staircase (ordered leading prefix). State + decision live together here (2026-07-23).
-private class SummaryDedup(private val active: Boolean) {
-    private val emittedParts = mutableListOf<String>()
+private class SummaryDedup(private val active: Boolean, shared: SharedSummaryParts? = null) {
+    // emittedParts + itemEmitted are TURN-scoped when shared is present (continuation rounds);
+    // recapCursor is deliberately per round — each round's stream re-starts its leading recap
+    // at position 0 of the shared list, which is exactly what suppresses it.
+    private val emittedParts = shared?.emittedParts ?: mutableListOf()
```
marcospaulo: **[nit]** The turn-scoped split is the right shape for the stated goal (detailed summary on every round, zero duplicates), and the PR description already names the accepted edge, so this comment is only about making that acceptance visible where the mechanism lives.

`ResponsesStreamTranslator.kt:112`:
```kotlin
private val itemEmitted = shared?.itemEmitted ?: mutableMapOf()
```

`itemEmitted` is keyed by the upstream `output_index`, and the backend restarts `output_index` at 0 on every continuation round (the new `SummaryTurnDedupTest` round 2 reuses slot 0). That makes the per-item set effectively turn-global per slot: a genuinely new round-2 reasoning item that lands in an already-used slot and happens to share a byte-identical non-leading paragraph with the round-1 item of that slot is suppressed at `val fresh = itemEmitted.getOrPut(oi) { HashSet() }.add(part)` (line 127), even though the two items are distinct. This is exactly the trade the PR description accepts ("overwhelmingly a restatement in practice, and the no-duplicates requirement wins"), so this is not a request to change the behavior.

One direction I would explicitly not take is re-keying the set by round or item identity: when the backend re-titles a repeat into the same slot, that keying is what lets the 2026-07-26 mirror duplication back in. What seems missing is the contract at the point of decision. The class comment just above (lines 96-107) documents the 2026-07-20 incident, where a turn-global SET over-suppressed a paragraph two distinct items shared, and the per-item split reads as the guard against it. A future reader sees that warning, then sees a turn-scoped map keyed by a per-round index, with nothing reconciling the two. Something in the existing dated-incident register, for example:

```kotlin
// The per-item sets are TURN-scoped per output_index slot (shared across continuation rounds,
// whose output_index numbering restarts at 0). A byte-identical non-leading part shared by two
// genuinely different items in the same slot is suppressed across rounds: accepted 2026-07-26,
// an exact identity is overwhelmingly a restatement and no-duplicates wins (operator call).
```

Preference, not requirement: if you judge the PR description to be enough, feel free to close this. The code comment is what survives the merge, which is why I raised it here.

marcospaulo (commented):
**Review summary** (pipeline: mechanical ast-grep + sharded lens review + evidence-gated filter; compact mode)

**Intent:** keep the detailed reasoning-summary request on every continuation round (reverting the #57 strip) while killing the duplicate sections that re-request causes, by hoisting summary-dedup state from per-round translators to a turn-scoped SharedSummaryParts on TurnMeta.

**Mechanical scan:** ran (72 rules: bundled kotlin pack + the repo's own `.rules` pack), 0 hits on added lines.

**Findings:** 4 raw -> 2 survived (1 medium, 1 nit), 2 dissolved at the delta gate (training-prior comments with no referent in this repo: a data-class ABI-break claim, but nothing is published as a binary artifact; an encapsulation prior on a deliberately shared coordination holder). Rank order below: the coverage-gap comment first (coupling degree 4), then the design-tension nit.

**Verified during review:** the turn-scoping claim holds in production wiring today. `TurnDriver.kt:288` stores `built.meta` on the drive and `:464` passes the same instance to `provider.streamTranslator` on every round (fold, re-anchor, tool_search), so `summaryParts` is genuinely shared across continuation rounds.

**Checked clean:** security lens (no injection/credential surface in the diff), flow-map pass over the shared-state seam (ownership and lifecycle of SharedSummaryParts are consistent; rounds are sequential, no concurrent mutation path found), ResponsesFold/ResponsesProvider skim, and the test files. `ResponsesStreamTranslatorTest.kt` was covered by the mechanical scan only (dispatch tier: mechanical-only).

**Flow map:** `flow-map.md` artifacts exist in the review workdir if you want the energy/state maps.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesProvider.kt
marcospaulo: Addressed in af57b0a — and you were right that the gap was real, not theoretical.

I went behavioral rather than `assertSame` on the meta instances: a new mock scenario `foldsummary` where the continuation round re-titles round 1's section verbatim (`output_index` restarting at 0, as the backend does) and then adds a genuinely-new one, driven through the real head in `HeadServerFoldTest` (HeadServer → TurnDriver → per-round translator). It asserts the re-titled section reaches the client exactly once and the new one still lands.

Reference identity is the mechanism; this pins its consequence, which also survives ways of breaking the sharing that an `assertSame` wouldn't see.

Red-proof, exactly your falsifiability scenario: swapping `summaryPartsShared = meta.summaryParts` for a per-round `SharedSummaryParts()` in `ResponsesProvider` turns **only** this test red — all 166 `dialect-openai-responses` tests stay green. That is the blind spot, now closed. Full gate green.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesStreamTranslator.kt
marcospaulo: Taken — the code comment is what survives the merge, and the unreconciled reading you describe (2026-07-20 warning above, turn-scoped map keyed by a per-round index below) is exactly what a future reader would hit. Added to the dated-incident register in af57b0a:

```kotlin
// The per-item sets are TURN-scoped per output_index SLOT (2026-07-26): continuation rounds share
// them and the backend restarts output_index at 0 each round, so a byte-identical non-leading part
// shared by two genuinely different items landing in the same slot is suppressed across rounds.
// Accepted (operator call): an exact identity is overwhelmingly a restatement, and no-duplicates
// wins. Re-keying the sets by round or item id is the one fix NOT to make — when the backend
// re-titles a repeat into the same slot, that keying is what lets the mirror duplication back in.
// The 2026-07-20 case stays intact: WITHIN a round, distinct items still get distinct slots.
```

I kept your "do not re-key by round or item identity" note in the comment rather than only in the PR description — it is the trap the next person fixing this would walk into. Behavior unchanged.

marcospaulo (commented):
Multi-lens review (5 lenses + seam flow-map over 22 files, sonnet shards; evidence-gated by an opus filter; mechanical scan: 82 ast-grep rules run, 0 skipped, 9 hits, all 9 dissolved as false positives or duplicates on named evidence).

Coverage disclosure: AGENTS.md, ResponsesFoldTest.kt and ResponsesStreamTranslatorTest.kt were tiered mechanical-only (rule coverage, no lens tokens) by the risk dispatch. Everything else received full lens context. A seam flow-map ran on the continuation-round/early-open state path; its two maps are in the review workdir (flow-map.md) and its verdicts fed the filter.

Scope note for the author: the PR body describes only the summary dedup. The diff also carries per-head knob overrides (8a1f431) and the early-open stream handoff (603b138). Both were reviewed; consider naming them in the description so the next reviewer does not discover them by diff archaeology.

No antibody drafts this round: the dissolved rule hits are prune signals on existing test rules (they misfire on real-socket integration tests), not new mechanically-expressible classes.

Review of: Revert the #57 continuation-summary strip and hoist reasoning-summary dedup to turn scope via SharedSummaryParts so exact-repeat sections never reach the client twice; the branch ALSO carries two undescribed features: per-head config overrides ([heads.<key>.overrides]) and an early-open stream handoff perf fix (603b138).

Findings: 1 high, 3 medium, 1 nit. (13 raw findings dissolved by evidence gates.)

Checked clean: api-abi, flowmap, mechanical.

Additional observations:
- [2026-07-26] security: Early-open SSE handoff: the added line `drive.emitter.ensureStarted()` at gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt:454 opens the client stream immediately after a 2xx upstream response. The inline comment claims this block runs only on `UpstreamClient.attemptRequest` success and that pre-stream failures still write their error frame first. The skim cannot verify `UpstreamClient.attemptRequest` semantics; confirm that 'success' means upstream authentication/authorization is fully complete before the 2xx, otherwise the client could observe a message_start frame before the upstream is authorized.
- [2026-07-26] security: Per-head TOML overrides: head keys and override maps flow from `topology.heads.mapValues { (_, head) -> head.overrides }` into `ConfigService.perHeadOverrides` and are keyed by `headKey` in `mergedRaw`. The skim does not show the TOML parser or `coerceAll`; verify that head keys are not user-controlled and that override values are coerced/sanitized so that path/command injection and unexpected knob values cannot be introduced via the `[heads.<key>.overrides]` layer.
- [2026-07-26] security: Upstream-derived content in error messages: `UpstreamFailureClassifier.quotaSuffix` appends the `plan_type` and reset timing fields from upstream JSON to the classified failure message, which is then sent to SSE error frames, non-stream envelopes, and daemon.log. These fields are not credentials, but they are upstream-controlled strings propagated into client-facing and log outputs; confirm downstream framing/escaping handles embedded control characters (e.g., newlines) to avoid log-injection or SSE-framing issues.
- [2026-07-26] build-config: Verified the override keys (maxInflight, upstreamTimeoutMs) match the Knob enum keys accepted by ConfigService.coerceAll.
- [2026-07-26] build-config: Verified the precedence chain documented in the comment matches ConfigService.mergedRaw ordering: defaults -> headOverrides -> perHeadOverrides -> fileLayer -> envLayer -> runtimeLayer.
- [2026-07-26] build-config: Verified the override values are strings, consistent with ConfigService.perHeadOverrides: Map<String, Map<String, String>>.
- [2026-07-26] build-config: No Gradle/CI/build-file changes were in this shard; review focused on config/hook consistency of the example TOML.
- [2026-07-26] correctness: ConfigService.layers() and PatchResult.effective only return the global config view; the new per-head override layer is not exposed in diagnostics or PATCH responses. If the control plane or dashboard uses these APIs, operators cannot verify per-head effective configs.
- [2026-07-26] correctness: The SharedSummaryParts lifecycle is implicit: it is created in ResponsesRequestBuilder.build and reused only because continuationRequest bypasses the builder. This is correct for the intended fold/reanchor paths, but it is a fragile contract that is not type-checked.
- [2026-07-26] correctness: Mechanical scan flagged UpstreamFailureClassifier.humanizeDuration line 120 for an 'else' branch in a when. The when is on Long ranges, not a sealed type, and the else is required for the seconds fallback. No action needed.
- [2026-07-26] tests: Daemon.kt is explicitly named as a source file whose diff adds logic (per-head ConfigService integration, providerContext, per-head InflightGate) with no matching test change.
- [2026-07-26] tests: TurnDriver.kt early-open handoff (ensureStarted) is covered on the happy path by HeadServerReviewTest, but the failure-after-open branch is not covered.
- [2026-07-26] tests: SummaryTurnDedupTest's bug-pinned case is valid: it would fail if SharedSummaryParts state were accidentally global. It does not by itself prove production wiring; HeadServerFoldTest covers that wiring end-to-end.
- [2026-07-26] tests: ResponsesFoldTest pins 'continuation keeps the summary request' as claimed in the PR intent.
- [2026-07-26] tests: Mechanical scan already flagged Thread.sleep, runBlocking, Dispatchers.IO, println, and sealed-over-stringly in the test and mock files; those are quality/style issues already recorded by the deterministic scan.
- [2026-07-26] flowmap: SseEmitter.started (line 58, not added in this PR) is a plain `var`; no @Volatile or AtomicBoolean. The early-open path now relies on ensureStarted() being idempotent across continuation rounds. Current access is sequential within the round coroutine, but the flag is public via hasStarted and is not thread-safe by construction.
- [2026-07-26] flowmap: Mechanical scan hit 'no-else-in-sealed-when' at UpstreamFailureClassifier.kt:120 is a false positive for this seed: the when is on Long range comparisons inside humanizeDuration, not on a sealed type.
- [2026-07-26] flowmap: CONTENT_FRAMES_OUT (TurnDriver.kt:696-700) excludes message_start and ping event frames; the keepalive comment ping (': ping\n\n') bypasses the emitter callback and is not counted, so pre-content tear reissue remains correctly enabled after early-open.
- [2026-07-26] flowmap: The early-open failure path is handled: if a continuation round fails after ensureStarted() but before content, the client sees message_start + ping + event:error (or abandon if client gone). Terminal is emitted by pipeline.finishStream -> SseEmitter.emitError.
- [2026-07-26] flowmap: Daemon.kt and ConfigService.kt changes are reachable from the seed only through provider construction / per-head watchdog timeouts; they introduce no per-turn shared state object owned by this seam.
- [2026-07-26] RELIEF_AUDIT 2026-07-26: scanned all 13 dissolutions; none re-examined. F017/F018 dissolved on named contrary code (per-test CountDownLatch; BufferingWireSink live passthrough fetched at head SHA). F001-F003/F007-F009 rule hits misfire on real-socket integration tests (lambda prune signal on kt-no-runblocking/kt-no-thread-sleep/tests-no-* packs). F010 dissolved as application-not-library; stands unless core is ever published.

gateway/app/src/main/kotlin/splice/app/Daemon.kt
```diff
@@ -666,9 +677,14 @@ public class Daemon(
...
                 inferenceToken = mgmtKey.get(),
+                // Re-read per head on EVERY admission (still hot-resizable): the ceiling belongs to
+                // the upstream ACCOUNT, not the gateway. One shared value meant a workflow fan-out
+                // admitted 100 concurrent streams into a single account, 429'd, and armed the shared
+                // cooldown — measured 67% turn failure at inflight=100 vs 0.3% at <=14 (perf jsonl,
+                // 2026-07-24). Per-head lets a slow upstream sit low while a fast one stays high.
                 gate = InflightGate(
-                    maxInflight = { config.getConfig().maxInflight },
-                    maxQueued = { config.getConfig().maxQueued },
+                    maxInflight = { config.getConfig(key).maxInflight },
```
marcospaulo, line 688: **[concern]** The per-head plumbing reads clean: `providerContext` resolves every per-head input from `config.getConfig(key)`, and the gate re-reads it on every admission, which keeps the hot-resize behavior the comment above this block describes. What is missing is a pin on this wiring. `ConfigServiceTest` covers the merge layer in isolation, but nothing fails if a future edit drops the key here:

```kotlin
gate = InflightGate(
    maxInflight = { config.getConfig(key).maxInflight },
    maxQueued = { config.getConfig(key).maxQueued },
),
```

`config.getConfig()` (no key) compiles in place of `config.getConfig(key)` at every one of these call sites, keeps the whole suite green, and silently restores the failure mode the comment at line 680 quantifies (one shared ceiling across all heads, measured 67% turn failure at shared inflight=100). The unit tests cannot catch it because they exercise the service, not the call sites.

Direction: one test with two heads carrying different `[heads.<key>.overrides]`, asserting the resolved per-head values differ. If driving the full Daemon is heavy, calling `providerContext(keyA, ...)` and `providerContext(keyB, ...)` and asserting the two `ProviderBuild.watchdog` budgets disagree is enough. That test fails the day the key argument is dropped or the override map stops reaching the `ConfigService` constructor at Daemon.kt:262.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt
```diff
@@ -442,6 +446,12 @@ internal class TurnDriver(
             amendBodyOnFailure = provider::amendBodyOnFailure,
         ) { resp ->
             drive.slot.touch()
+            // Upstream answered 2xx and the stream is ours — open the turn on the wire immediately
+            // instead of waiting for the first content block. This block runs ONLY on success (see
+            // UpstreamClient.attemptRequest), so a pre-stream failure still writes its error frame
+            // first and nothing here pre-empts it. Recovers p50 2840ms of frozen screen per codex
+            // turn; also gives the keepalive pinger an opened stream to ping into.
+            drive.emitter.ensureStarted()
```
marcospaulo, line 454: **[suggestion]** The early-open handoff is well covered where the turn succeeds: the holdstart test proves `message_start` reaches the client while upstream is still silent, and the existing tear test already drives the failure side of this same path (the mock commits 200 + headers, `ensureStarted()` fires here at handoff, then the socket drops before any content).

The gap is in what the tear test asserts. It pins error honesty (exactly one `event: error`, `overloaded_error`, no `message_stop`) but not the frame contract this change creates on the failure path: the client now sees `message_start` and `ping` before the error frame. The current assertions stay green if those early frames disappeared from the torn path entirely, as long as the error frame survives. One line in the existing tear test closes it, and no new mock scenario is needed since "tear" already produces exactly this sequence:

```kotlin
assertTrue(
    body.indexOf("event: message_start") in 0 until body.indexOf("event: error"),
    "message_start must precede the error frame on the early-open failure path: $body",
)
```

That turns the ordering you established into something a regression has to break loudly.

config/splice.example.toml
```diff
@@ -191,3 +191,19 @@ discovery_prefix = "claude-kimi--"
 pinned_model = "k3[1m]"
 [heads.claude-kimi.claude]
 command = "claude-kimi"
+# PER-HEAD KNOBS. Same names as [defaults], same string values — but scoped to THIS head, so a
```
marcospaulo: **[suggestion]** This block is the right operator doc in the right place: the maxInflight framing (the concurrency the upstream account tolerates, not what the gateway can hold) matches the measured 429-storm story in the Daemon comment, and the precedence chain matches `mergedRaw` in ConfigService. One seam: the block leans on `[defaults]` three times ("Same names as [defaults]" here, the precedence line, and "keep the long cap HERE instead of [defaults]" at line 207), but this 209-line example never defines a `[defaults]` section. An operator copying this file cannot enumerate the valid knob names or see where global defaults live without leaving the file for AGENTS.md.

Cheapest closure is a short section above the per-head block:

```toml
[defaults]
maxInflight = "8"
upstreamTimeoutMs = "600000"
```

Or, if keeping the example lean is deliberate, reword the comment to name the source of truth directly (the Knob enum keys in ConfigService) so the file stays self-contained. Docs only; the wiring and precedence themselves check out.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
```diff
@@ -86,6 +93,34 @@ public object UpstreamFailureClassifier {
...
+            ?: err["resets_at"]?.jsonPrimitive?.longOrNull?.let { it - System.currentTimeMillis() / MS_PER_S }
+        resetsInS?.takeIf { it > 0 }?.let { parts += "resets in ${humanizeDuration(it)}" }
+        return if (parts.isEmpty()) "" else " (${parts.joinToString(", ")})"
+    }
+
+    /** Seconds -> "6d 4h" / "4h 12m" / "45s". Coarse on purpose: the operator needs the ORDER of
+     *  magnitude ("six days, switch heads") far more than the exact second. */
+    private fun humanizeDuration(totalSeconds: Long): String {
+        val d = totalSeconds / SECONDS_PER_DAY
+        val h = (totalSeconds % SECONDS_PER_DAY) / SECONDS_PER_HOUR
```
marcospaulo, line 114: **[suggestion]** The new quota-suffix tests cover the `resets_at` arm with plan_type and the no-fields case, and `humanizeDuration`'s coarse buckets are the right resolution for the operator decision ("six days out, switch heads"). One branch remains unpinned, and it is the precedence arm (UpstreamFailureClassifier.kt:104-105):

```kotlin
val resetsInS = err["resets_in_seconds"]?.jsonPrimitive?.longOrNull
    ?: err["resets_at"]?.jsonPrimitive?.longOrNull?.let { it - System.currentTimeMillis() / MS_PER_S }
```

When both fields are present, `resets_in_seconds` wins, and it is the arm with no test. A typo in that key or a unit slip stays green. Honest uncertainty: the diff does not show which field the live ChatGPT quota payload actually carries, so I cannot tell whether the untested arm is the production path or a defensive fallback. A case mirroring the existing one closes it either way:

```kotlin
val body = """{"error":{"type":"usage_limit_reached","message":"The usage limit has been reached","resets_in_seconds":3600}}"""
assertTrue(http(body, 429).message.contains("resets in 1h 0m"))
```

gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
```diff
@@ -131,8 +131,21 @@ public data class TurnMeta(
...
+     *  build fresh translators; without a shared set, a section re-titled by a continuation round
+     *  passes each round's per-instance dedup and lands as a duplicate — the 2026-07-26 mirror
+     *  duplication). Null on dialects that don't render reasoning summaries. */
+    val summaryParts: SharedSummaryParts? = null,
 )
 
+/** The shared state behind TurnMeta.summaryParts: the ordered parts already emitted to the
+ *  client this turn, plus the per-item exact set the dedup's within-item arm matches against.
+ *  Mutable per-turn coordination, never compared by value. */
+public class SharedSummaryParts {
```
marcospaulo, line 162: **[nit]** The turn-scoped hoist is the right shape, and the lifecycle works precisely because `continuationRequest` bypasses the builder and reuses one `TurnMeta`. One contract worth a sentence on the new type (Turn.kt:144):

```kotlin
public class SharedSummaryParts {
    public val emittedParts: MutableList<String> = mutableListOf()
    public val itemEmitted: MutableMap<Int, MutableSet<String>> = mutableMapOf()
}
```

Correctness rides on a single-writer-per-round discipline: each round's translator mutates these from one coroutine, and the FoldRunner/ReanchorRunner loops run rounds strictly sequentially. The current doc ("Mutable per-turn coordination, never compared by value") covers scope but not that access discipline, so the next person to touch the round loops has to rediscover the invariant. Two lines of KDoc keep it where the next edit happens, e.g. "Mutated by one round's translator at a time; rounds run strictly sequentially in the fold/reanchor loops, so the absence of synchronization is deliberate".

Preference, not requirement: I am not asking for synchronization (there is no concurrent access today; the flow is sequential by loop construction), and `internal` would be wrong here since the dialect module reads these across a module boundary.

gateway/app/src/main/kotlin/splice/app/Daemon.kt
marcospaulo: Addressed in 61d83de. You were right that the unit tests could not reach this — and the seam is worse than "heavy to drive": nothing public exposes per-head resolved config at all (`configJson()` at ControlServer.kt:467 renders `config.getConfig()`, the global view, by design for the dashboard).

So I took your fallback and made it reachable: `providerContext` and `ProviderBuild` go `private` → `internal` (visibility only, no logic change, no new production caller). New `DaemonPerHeadConfigTest` builds a two-head topology differing only in `[heads.<key>.overrides]` and asserts the resolved `watchdog.totalCap` and `maxInflight` disagree.

Red-proof: changing `config.getConfig(key)` to `config.getConfig()` at Daemon.kt:350 fails it. Reverted after.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt
marcospaulo: Added in 61d83de, as your one line in the existing tear test — no new mock scenario needed, exactly as you said.

Red-proof: removing `drive.emitter.ensureStarted()` at TurnDriver.kt:454 fails it with precisely the intended message (`message_start must precede the error frame on the early-open failure path: event: error ...`), and separately times out the holdstart test. So the ordering is now something a regression has to break loudly, on both the success and failure sides.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
marcospaulo: Added in 61d83de, with one change to your sketch. You flagged the precedence arm as the untested one, so the new case sends **both** fields rather than `resets_in_seconds` alone: `resets_in_seconds: 3600` plus a `resets_at` six days out. It asserts `resets in 1h 0m` **and** that `6d` does not appear, so a fallthrough cannot pass by accident.

Red-proof: swapping the `?:` order at UpstreamFailureClassifier.kt:104-105 fails it.

On your honest-uncertainty note — the diff does not settle which field the live payload carries either, and I did not probe production to find out. The test pins the documented precedence regardless of which arm is the hot one.

gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
marcospaulo: Added in 61d83de. I verified the invariant rather than restating it: `FoldRunner.run` is a plain `while (true) { postRound(...) }` with no `launch`/`async` around a round, so rounds genuinely cannot overlap and the missing synchronization is deliberate.

The KDoc now says that, and adds the forward-looking half — a future round loop that overlaps rounds must add synchronization before sharing this. Kept `public` for the reason you gave (the dialect module reads it across a module boundary); no synchronization added.

config/splice.example.toml
marcospaulo: Addressed in 61d83de — I took your second option (reword) rather than adding a `[defaults]` table, for a reason worth stating since you offered both.

A live `[defaults]` block in the example changes what a copier actually gets: those knobs would then govern **every** head, which is the exact failure the per-head layer exists to solve, and the comment 13 lines below already tells operators to keep the long cap out of `[defaults]`. The example would have been arguing with itself.

The block now names the source of truth directly — the Knob enum keys (`maxInflight`, `maxQueued`, `upstreamTimeoutMs`, `streamIdleMs`, `firstByteTimeoutMs`, all five verified present in Knob.kt), notes they are always strings, and says `[defaults]` is an optional top-level table this example deliberately omits, with why. The file is self-contained now without leaving for AGENTS.md.

`ExampleConfigTest` still parses it green.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnDriver.kt
marcospaulo: **Correction to my reply above.** The assertion I described there was a flake, and it turned CI red (`report` job, `HeadServerReviewTest`, 115 tests / 1 failed). Recording it here because that reply currently reads as a clean close and it was not.

The red-proof I ran *was* real — removing `ensureStarted()` did fail it with the intended message — but a passing red-proof does not make the assertion correct. It demanded `message_start` be **present** on the torn path, and presence depends on where the premature EOF surfaces: during upstream response setup (no handoff, `ensureStarted` never runs) or during the body stream (handoff happened). That is environment timing, not a contract. It passed 3/3 locally and lost the coin flip on CI.

`24eefbf` pins the invariant instead — `message_start` must never arrive **after** the error frame:

```kotlin
val startAt = body.indexOf("event: message_start")
assertTrue(startAt == -1 || startAt < body.indexOf("event: error"), ...)
```

Both legitimate timings pass; the regression you were guarding against (early frames moving after the error) still fails. The **presence** half stays pinned deterministically by the holdstart test at `HeadServerReviewTest.kt:137`, where the mock blocks before emitting anything so the handoff is guaranteed. So your contract is still covered — it just needed the deterministic scenario for presence and the torn one for ordering.

Verified: 3 consecutive full runs of the class green, full gate PASS, and `report` green on `24eefbf`.

.claude/hooks/orchestrator.py
```diff
@@ -149,6 +150,21 @@ def stop(data: dict) -> int:
...
+def _walls_grant_active() -> bool:
+    """True while an operator-issued /grant is live (see modules/userpromptsubmit/03_grant_command.py).
+
+    A grant is issuable ONLY from a UserPromptSubmit hook, which fires only on text a human typed
+    into the prompt box — an assistant emits tool calls, never a user prompt. So this stays
+    operator-only by construction, not by policy. Expiry is enforced HERE, at the gate, so a stale
+    file can never hold the wall open; the grant module reports the same rule.
+    """
+    try:
+        raw = json.loads((ROOT / ".claude/state/walls-grant.json").read_text(encoding="utf-8"))
```
marcospaulo: **[blocker]** The grant channel is well engineered on every path it controls: CLI mode is read-only by design, the installer asserts its patches instead of assuming them, the self-test proves red then green end to end, and expiry is enforced at the gate rather than trusted to the writer. The problem is the one path it does not control: the file the gate trusts.

The docstring on `_walls_grant_active()` claims the grant is 'operator-only by construction'. As implemented, that claim fails in one step: an assistant can mint a grant itself. The chain, each link verified in the tree:

1. `WALL_PATHS` (orchestrator.py:54) is `(".rules", ".claude/hooks", ".claude/settings.json", "sgconfig.yml")`. `.claude/state/` is not among them, so a normal Write of `.claude/state/walls-grant.json` passes the guard at line 102 and matches no ast-grep rule.
2. `_walls_grant_active()` (orchestrator.py:162-163) reads that file and trusts the `until` field with no provenance check.
3. The next write to a wall path hits line 102 with `not _walls_grant_active()` false, so the block never fires. The wall is open.
4. install.sh:113 already demonstrates the forge: its own self-test writes the grant file directly with python and the gate goes ALLOW. `.gitignore:61` keeps the forged file out of git, so the bypass leaves no trace.

This fails whenever an agent decides the wall is in its way, because the trust anchor is a file the agent can write. Two fixes, and both are needed:

1. Add `.claude/state/walls-grant.json` to `WALL_PATHS` so the Write/Edit path is covered. Note this does not cover Bash: `echo '...' > .claude/state/walls-grant.json` never goes through PreToolUse, which is why the second fix is the real one.
2. Sign the record. Have the UserPromptSubmit module HMAC the payload with a key held outside the repo (for example in the system keyring, provisioned by install.sh), and have the gate verify the signature before trusting `until`. Include a session identifier in the signed payload so the audit record answers who issued the grant and from where, not only when it expires. A signature check also makes a forged or hand-extended record tamper-evident instead of silently accepted.

Until fix 2 lands, the channel is operator-only by policy, and the policy is one Write call thick.

dev/campaigns/proxy-hardening/oracle/capture.mjs
```diff
@@ -0,0 +1,213 @@
...
+      { host: '127.0.0.1', port, path: '/v1/messages', method: 'POST', headers: { 'Content-Type': 'application/json' } },
+      (r) => { let t = ''; r.on('data', (c) => { t += c; }); r.on('end', () => res({ status: r.statusCode, sse: t })); },
+    );
+    req.on('error', rej);
+    req.end(JSON.stringify(body));
+  });
+}
+
+async function main() {
+  const check = process.argv.includes('--check');
```
marcospaulo, line 197: **[concern]** Freezing the oracle while the legacy stack is still green is the right instinct, and the manifest's recorded sha plus the explicit exclusions list are exactly the provenance a frozen oracle needs. `--check` is where the drift tripwire should live, and today it cannot trip.

The docstring (capture.mjs:31) promises '--check (re-capture to temp, diff vs committed)'. The implementation (lines 149-211) does the first half only: `outDir` redirects to the temp dir, the temp tree is deleted at line 209, and the exit code at line 210 reflects scenario success alone. Nothing in a `--check` run reads the committed `fixtures/`, and `mockSha` is recomputed from the live `server/test/codex-proxy.test.mjs` on every run, so even mock drift is invisible to it. This is falsifiable in one step: change behavior in `server/`, run `node dev/campaigns/proxy-hardening/oracle/capture.mjs --check`, and it still exits 0 as long as no scenario errors.

That would be a doc nit if nothing consumed it, but it is wired as a proof: `oracle:check` in package.json, and the `verify` command of ledger item INF-03 (proxy-hardening.toml:845-847). The one thing keeping this below a blocker is that INF-03 is still `todo`, so the vacuous proof has not yet been used to verify anything. The campaign's own polarity law (a wall that passes while its item is unfinished reads vacuous) is the doctrine this violates in spirit.

Direction: after capture in check mode, diff `tmp/fixtures/*.json` against `fixtures/*.json` byte-for-byte (or compare the freshly computed per-scenario sha256 rows against the committed `_manifest.json`) before the `rmSync`, print the drifted scenario names, and exit nonzero on any mismatch. The docstring then tells the truth and INF-03's verify becomes a real gate.

dev/walls-grant/install.sh
```diff
@@ -0,0 +1,148 @@
...
+
+echo "== splice /grant installer =="
+echo "repo: $ROOT"
+
+# ---------------------------------------------------------------- 1. the module
+mkdir -p "$MODDIR"
+cp "$ROOT/dev/walls-grant/03_grant_command.py" "$MODDIR/03_grant_command.py"
+# The / menu surface. .claude/commands/ is gitignored in this repo (per-developer Claude Code
+# config is deliberately not shipped), so the canonical copy lives here and is installed locally.
+mkdir -p "$ROOT/.claude/commands"
```
marcospaulo, line 34: **[suggestion]** The installer's discipline is good where it aims: assert-on-patch instead of blind replace, a red-then-green self-test driven through the real gate, and the module itself landing under the walled `.claude/hooks/` tree. The menu surface does not get the same treatment.

Lines 26-28 copy `grant.command.md` and `install-walls.command.md` into `.claude/commands/`, which `.gitignore:51` excludes from version control, and `.claude/commands` is not in `WALL_PATHS` (orchestrator.py:54). So the file the operator's `/grant` keystroke actually executes is untracked and unwalled: an on-disk edit, accidental or otherwise, leaves no git trace and trips no gate, and the bang-line in that file runs whatever bash it names. The canonical copies in `dev/walls-grant/` are tracked, which is the right home; the gap is that installed copies are never reconciled against them.

To be fair about what this does and does not buy: it does not enlarge what an agent can do directly, since Bash already covers direct action. What it closes is silent divergence between the reviewed canonical file and what the operator actually runs.

Direction, two options: after the `cp`, record sha256 of both installed files (somewhere like `.claude/state/`) and have install re-runs, or the read-only `/grant` status path, verify the hashes and complain on divergence. Or symlink instead of copy so the tracked file is the single source, if Claude Code follows symlinks for command files; I have not verified that it does, so treat the symlink route as a quick experiment before choosing it.

gateway/app/src/test/kotlin/ExampleConfigTest.kt
```diff
@@ -61,4 +62,28 @@ class ExampleConfigTest {
...
+        val used = topology.heads.values.flatMap { it.overrides.keys }.toSet()
+        assertTrue(used.isNotEmpty(), "the example must keep demonstrating per-head overrides")
+        used.forEach { assertTrue(Knob.byKey.containsKey(it), "example sets unknown knob '$it'") }
+
+        // 2. Knob names the COMMENTS teach. Anything camelCase-and-backticked-or-listed in the
+        //    per-head doc block must resolve; that is what went stale and got hand-fixed once.
+        val docNames = Regex("\\b([a-z]+[A-Z][A-Za-z]*)\\b")
+            .findAll(toml.lines().filter { it.trimStart().startsWith("#") }.joinToString("\n"))
+            .map { it.groupValues[1] }
+            .filter { it.endsWith("Ms") || it.startsWith("max") }
```
marcospaulo: **[suggestion]** Having the example's knob vocabulary answer mechanically to the Knob enum is a real upgrade over the prose that rotted, and the two-part shape (live override tables plus comment-taught names) covers both ways the file can lie. Part 2 is narrower than its own contract.

The comment (lines 80-81) says anything camelCase in the doc block must resolve, but line 85 keeps only names that start with `max` or end with `Ms`:

```kotlin
.filter { it.endsWith("Ms") || it.startsWith("max") }
```

Today that happens to be lossless: the example's comments name only `firstByteTimeoutMs`, `maxInflight`, `maxQueued`, `streamIdleMs`, `upstreamTimeoutMs`, and the one dropped token (`xAI`) is not a knob. The check fails the first time a comment teaches a knob outside that shape: `toolSurface` (Knob.kt:133), `usageWarnPct` (Knob.kt:222), `pinnedModel`, `foldMarkerText`, and `contextWindowOverride` are all live keys the filter would silently drop, and silent skipping is exactly the drift this test exists to catch, per its own comment ('that is what went stale and got hand-fixed once').

Two directions, either closes it. Broaden the check: resolve every camelCase comment token against `Knob.byKey` and keep a small explicit allowlist for prose tokens like `xAI`, so a future unknown token fails loudly instead of being pre-filtered out of the assertion. Or narrow the comment to say the check covers the timeout-and-capacity family only. My preference is the first, since the enum is the authority and the allowlist keeps prose tokens visible, but that is a preference, not a requirement.

.claude/hooks/modules/userpromptsubmit/03_grant_command.py
```diff
@@ -0,0 +1,229 @@
...
+sys.path.insert(0, str(pathlib.Path(__file__).resolve().parent.parent.parent))
+from orchestrator.result import HookResult  # noqa: E402  # pyright: ignore[reportMissingImports]
+
+# Walls work is focused and short. A long grant is a grant someone forgot to revoke, so the
+# ceiling is deliberately tight — re-issuing is one line, an all-day open gate is a liability.
+MAX_MINUTES = 120
+DEFAULT_MINUTES = 30
+GRANT_REL = ".claude/state/walls-grant.json"
+
+WALL_PATHS_DOC = ".rules/, .claude/hooks/, .claude/settings.json, sgconfig.yml"
```
marcospaulo, line 73: **[suggestion]** The bounded lifetime and the mandatory reason are the right levers for this channel, and 120 minutes with a 30-minute default is a sensible ceiling. The breadth of what a grant opens is all-or-nothing.

The payload (03_grant_command.py:137-142) carries `until`, `minutes`, `reason`, `issued_at`, and no scope field, and `_walls_grant_active()` in orchestrator.py returns a bare boolean. So `/grant 30 fix one rule` opens `.rules/`, `.claude/hooks/`, `.claude/settings.json`, and `sgconfig.yml` simultaneously, even though walls work almost always touches exactly one of those. This fails to limit blast radius precisely in the common case: a grant issued for a narrow job leaves three unrelated wall paths writable for the rest of the window.

Direction: accept an optional path argument, `/grant 30 <reason> .rules`, record a `paths` list in the payload, and have the gate check the requested relpath against it before honoring the grant. This is a hardening suggestion, not a requirement, and I would sequence it after the forgeability fix at orchestrator.py:162 rather than instead of it: while the file is forgeable, a scope field only constrains honest issuers, since a forged payload could name every path.

.rules/kotlin-splice/kt-catch-swallows-cancellation.yml
```diff
@@ -12,5 +17,5 @@ rule:
   regex: ':\s*(kotlin\.)?(Exception|Throwable)\b'
   not:
     has:
-      regex: CancellationException
+      pattern: $_ is CancellationException
```
marcospaulo: **[suggestion]** The documented widening from a free-text regex to a structural type-check closed the embarrassing hole (a comment merely mentioning CancellationException satisfied the old exemption), and the header tells that story well. The rule stops one rung short of its own message.

The message (line 12) requires `if (e is CancellationException) throw e`. The exemption (lines 18-21) requires only the type-check:

```yaml
not:
  has:
    pattern: $_ is CancellationException
    stopBy: end
```

So this passes the wall while still swallowing cancellation:

```kotlin
catch (e: Exception) {
    if (e is CancellationException) { log.debug { "cancelled" } }
    recover()
}
```

It fails whenever an author type-checks without rethrowing, because `$_ is CancellationException` matches the `if` condition and the wall goes quiet while the watchdog abort or client disconnect gets eaten: the zombie-stream class the header names as the reason for the rule.

Direction: bind the checked variable and require its rethrow in the same catch block:

```yaml
not:
  all:
    - has:
        pattern: $V is CancellationException
        stopBy: end
    - has:
        pattern: throw $V
        stopBy: end
```

ast-grep binds `$V` consistently across the `all`, so a rethrow of some other variable does not satisfy the exemption. Worth a rule-test with the log-only body above as the red case and the standard `if (e is CancellationException) throw e` as the green one.

.rules/kotlin-splice/kt-no-silent-result-collapse.yml
```diff
@@ -24,12 +31,38 @@ files:
...
+        - has:
+            stopBy: end
+            regex: "\\brunCatching(Cancellable)?\\b"
+        - not:
+            has:
+              stopBy: end
+              regex: "\\bonFailure\\b"
+    # B. split form: `val r = runCatching { … }` then `r.getOrNull()` on a later line.
+    - all:
+        - kind: property_declaration
```
marcospaulo, line 69: **[suggestion]** Branch B is a well-told fix: the metavariable identity on `$RECV` is exactly what keeps it from firing on unrelated vals, and the header explains the subtree evasion it closes. Two shapes of the same collapse still walk through.

The branch is scoped to `kind: property_declaration` (line 48) and its preceding binding must be `val $RECV = ...` (line 61). Both of these escape while remaining the silent collapse the message bans:

```kotlin
var r = runCatching { read() }        // var binding, branch B requires val
val token = r.getOrNull()

fun f(): String? {
    val r = runCatching { read() }
    return r.getOrNull()              // collapse in a return, not a property initializer
}
```

Branch A cannot see either shape, because `runCatching` is not inside the collapse node's own subtree. So the class the wall claims (any silent Result collapse) is open on two textual paths.

Direction, in increasing order of risk: the `var` widening is mechanical, accept `var $RECV = $$$INIT` alongside the val pattern in the `follows` arm and add a red rule-test for it. The broader gap (collapse sites outside property initializers) means relaxing the `kind: property_declaration` anchor so a `$RECV.getOrNull()` anywhere after the binding is caught; I have not verified how ast-grep's `follows` scopes across statement boundaries inside a function body, and a too-loose version would flag every later read of a legitimately handled Result, so treat that arm as a sketch to validate with rule-tests before landing, not as a finished rule. Landing just the `var` arm already halves the hole.

.rules/rules/launcher-no-pkill.yml
```diff
@@ -12,7 +25,32 @@ note: |
...
 files:
-  - server/launcher/**/*.mjs
+  - server/**/*.mjs
 rule:
-  kind: string_fragment
-  regex: pkill
+  any:
+    - kind: string_fragment
+      regex: pkill
+    - kind: binary_expression
```
marcospaulo, line 46: **[suggestion]** The widening history in this file's header is the right kind of rule evolution, and scoping the split-literal branch to exec-family arguments already excludes the HTML concatenation it would otherwise flag. The branch is still wider than the bug class it names.

Lines 33-56 match any `+` whose two operands are string or template literals inside execSync/exec/execFileSync/spawn/spawnSync arguments, with no constraint on the operand text. The stated target is the split-evasion shape `'pk' + 'ill -f '`, but the pattern fires on any two-literal concatenation in those positions, for example `execSync('ls' + ' -la')`. (One correction to the example I would have reached for first: `execSync('pgrep -f ' + name)` does not match, because `name` is an identifier, not a literal.) The header's own verification, 0 hits on the codex-login.mjs HTML concat, held because that concat sits outside exec arguments, not because the branch is content-aware. So the over-breadth is latent today and becomes a false positive the first time someone legitimately joins two literals in a spawn call, and a wall that flags compliant code trains authors to reach for `ast-grep-ignore`, which weakens every wall including this one.

Direction: constrain the concatenation's text to the pkill morpheme so the branch targets the same class as the string_fragment branch it backs up:

```yaml
- kind: binary_expression
  regex: pk['"`]\s*\+\s*['"`]ill
  inside:
    stopBy: end
    kind: arguments
    inside:
      stopBy: neighbor
      kind: call_expression
      has:
        field: function
        regex: ^(execSync|exec|execFileSync|spawn|spawnSync)$
```

The regex matches the node text including its quotes, so ordinary spacing variants still match; template interpolation inside the fragments might need a second arm, which I have not verified, so treat that as the known edge to rule-test. A red/green rule-test pair, `'pk' + 'ill -f '` red and `'ls' + ' -la'` green, pins both sides of the fix.

dev/campaigns/proxy-hardening.toml
```diff
@@ -0,0 +1,871 @@
+# proxy-hardening.toml — never-fails / self-heals / just-works campaign for splice (2026-07-26).
+# LAW: concept #945 — this manifest IS the memory; access ONLY via dev/campaigns/manifest.py
+#   (`python3 dev/campaigns/manifest.py dev/campaigns/proxy-hardening.toml <verb>`). Raw edits skip
+#   the flock and the validate-rollback gate. NOTE: manifest.py's DEFAULT path argument is stale in
```
marcospaulo: **[suggestion]** The header doing its own honesty work, admitting the stale default instead of hiding it, is very much in the campaign's spirit. But the NOTE converts a one-line tool bug into a permanent reading requirement for every future session.

Lines 2-6 make CLI-only access the law and then document that the CLI's default ledger path is wrong: manifest.py:100 sets `DEFAULT = os.path.join(os.path.dirname(__file__), "kotlin-hardening.toml")`, and no such file exists in `dev/campaigns/` (the directory holds six other ledgers). So a bare, fully lawful `python3 dev/campaigns/manifest.py selftest` ends in a file-not-found traceback, and the only thing between an amnesiac session and that traceback is having read this NOTE first. An instruction a session must remember is precisely the failure mode the ledger design exists to remove.

Direction: change the default to `proxy-hardening.toml`, or derive it (for example, the single `*.toml` in `dev/campaigns/` carrying a campaign header). That is one line in manifest.py. If manifest.py is vendored from torad-fleet and you would rather not diverge it, the fallback is a two-line wrapper in this repo that always passes the path explicitly, so the lawful invocation works without memorizing the NOTE. Either way, the NOTE can then shrink to a pointer instead of a workaround.

gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
```diff
@@ -131,8 +131,33 @@ public data class TurnMeta(
...
+ *  client this turn, plus the per-item exact set the dedup's within-item arm matches against.
+ *  Mutable per-turn coordination, never compared by value.
+ *
+ *  ACCESS DISCIPLINE (2026-07-26 review): mutated by ONE round's translator at a time. The
+ *  fold/re-anchor/tool_search loops drive rounds strictly sequentially (`FoldRunner.run` is a
+ *  plain `while (true) { postRound(...) }` — no launch/async around a round), so the absence of
+ *  synchronization here is deliberate, not an oversight. A future round loop that overlaps rounds
+ *  must add synchronization before sharing this. Public, not internal: the dialect module reads
+ *  these across a module boundary. */
+public class SharedSummaryParts {
```
marcospaulo, line 162: **[suggestion]** The KDoc does the thing comments rarely do: it names the discipline, names the evidence for it (sequential rounds in `FoldRunner.run`), and names the condition under which the design must change. What a comment cannot do is enforce any of that.

`SharedSummaryParts` exposes `MutableList` and `MutableMap` as public vals (Turn.kt:156-159). The one real consumer I found (ResponsesStreamTranslator.kt:119-135) uses a narrow surface: an indexed read (`emittedParts[cursor]`), `add`, `getOrPut(...).add(...)`, and the `indices` bound. That indexed read is the concrete coupling: the translator's cursor logic trusts the list's order and length, and any in-repo caller can `clear()`, reorder, or truncate the list today with no compile error and no test to catch it.

To be explicit about status: this is a preference, not a requirement. Nothing misuses the class now, the documented discipline is honest, and the public visibility across the module boundary is justified. If you want the discipline mechanical rather than documentary, the class can keep the collections private and expose roughly the four operations the dialect actually performs:

```kotlin
public class SharedSummaryParts {
    private val emittedParts = mutableListOf<String>()
    private val itemEmitted = mutableMapOf<Int, MutableSet<String>>()

    public fun partAt(cursor: Int): String? = emittedParts.getOrNull(cursor)
    public val emittedCount: Int get() = emittedParts.size

    /** Returns false when the part was already emitted (a dedup hit). */
    public fun markEmitted(outputIndex: Int, part: String): Boolean {
        val fresh = itemEmitted.getOrPut(outputIndex) { mutableSetOf() }.add(part)
        if (fresh) emittedParts.add(part)
        return fresh
    }
}
```

That keeps the class public for the cross-module boundary while making 'append-only, dedup-checked mutation' a property of the type instead of a property of the comment. If the answer is that the KDoc is enough for now, that is a defensible call; the risk only materializes when a second mutator appears.

.claude/hooks/modules/userpromptsubmit/03_grant_command.py
```diff
@@ -0,0 +1,229 @@
...
+    except OSError:
+        pass
+    return "§grant — REVOKED. Wall paths are blocked again." if existed else "§grant — nothing to revoke."
+
+
+def applies(data: dict) -> bool:
+    # Defensive: this module is operator-only BECAUSE it is UserPromptSubmit-only. If a future
+    # refactor globs it into another lifecycle, refuse rather than silently become grantable
+    # by an assistant. The runner passes the lifecycle through when it knows it.
+    if data.get("hook_event_name") not in (None, "UserPromptSubmit"):
```
marcospaulo, line 254: **[nit]** The defensive instinct here is right, and the comment names exactly what it defends against: a future refactor globbing this module into another lifecycle, converting the grant channel from structure into suggestion. One arm of the check is quieter than that intent.

`if data.get("hook_event_name") not in (None, "UserPromptSubmit")` (line 169) admits a payload that omits the field entirely. I traced how the module is actually reached: runner.py (lines 111-119) passes the raw stdin payload through without injecting a lifecycle, and the module's placement under `modules/userpromptsubmit/` plus the `prompt`-is-a-string gate two lines below already exclude every non-prompt lifecycle. So the None arm only matters for a prompt-carrying payload that lacks the field, or for a direct hand-rolled invocation, and that second case is dominated by the state-file issue at orchestrator.py:162.

Requiring the explicit value would make the tripwire louder, but it would break `/grant` outright on any Claude Code payload shape that omits `hook_event_name`, and I cannot confirm from this repo which shapes the CLI has shipped. Marking that as genuine uncertainty rather than guessing: if the field has always been present for UserPromptSubmit, tightening to `!= "UserPromptSubmit"` makes the check do what the comment advertises; if not, the current shape is the right compromise and a one-line comment saying so saves the next reader this same trace. Fine to close with either resolution.

dev/campaigns/proxy-hardening/walls/wall_registry.toml
```diff
@@ -0,0 +1,453 @@
...
+#
+# LAW #924 / #954: an item with no wall is UNENFORCED DOCUMENTATION. Every row below is
+# checked by dev/campaigns/proxy-hardening/walls/campaign_wall_gate.py in BOTH directions, and a row may NEVER
+# be deleted to silence the gate — the fix is BUILDING THE WALL.
+#
+# THE POLARITY LAW: a wall whose item is todo/in_flight MUST FAIL. A wall that passes while
+# its item is unfinished is VACUOUS and reads RED. You cannot land a wall that was never red,
+# and you cannot mark an item done until its wall flips. Green is EARNED, never protected.
+#
+# BIRTH STATE 2026-07-26: 83 items, 80 UNWALLED, 3 walled-and-correctly-red.
```
marcospaulo: **[nit]** The polarity law and the 'honest work inventory' framing make this registry a real enforcement census rather than decoration, which is exactly right for a campaign file. The census header has drifted from the census.

Line 11 records '83 items, 80 UNWALLED, 3 walled-and-correctly-red'. The file below it now holds 88 `[[item]]` rows and 7 non-empty `wall =` paths (counted mechanically). In most files a stale count is cosmetic; in a file whose job is to be the number auditors and the campaign gate trust, the header is the one place that cannot drift, because a reader will quote 83/3 from the top and never recount the rows.

Two cheap fixes, either works: update the header to 88/81/7 (81 unwalled), or drop the numbers from the header and point at the rows ('the count is the rows; do not trust any summary, including this one'). The second option cannot rot.

marcospaulo (commented):
ROUND 2 review, head 24eefbf. Round 1 (head 603b138) posted 5 findings; all five were verified fixed in 61d83de + 24eefbf before this run, so the round-1 Kotlin surface got a drift check (clean) rather than a re-review. The new mass reviewed this round: the ast-grep walls audit, the /grant operator channel, the campaign ledger and oracle, and SharedSummaryParts' new non-null default.

Method: 5 lenses + a seam flow-map over 129 files (65 deep, 70 skim, 31 mechanical-only, disclosed), sonnet shards, opus evidence filter. Mechanical scan: 108 ast-grep rules run, 0 skipped, 9 hits, all 9 being round-1 adjudicated false positives recurring unchanged (dissolved again with the same named evidence). One disclosed tier exception: the security lens took the 229-line /grant module deep instead of digest-only, because a wall-bypass channel outguns the risk signals the dispatcher tiers on.

Review of: ROUND-2 review of PR #58 at 24eefbf. Four payloads now share the branch: (1) the round-1 subjects (summary dedup, per-head overrides, early-open handoff),already reviewed at 603b138 and fixes verified in 61d83de/24eefbf, regression-check only; (2) the ast-grep walls audit (33->36 walls, 29 rule improvements),the repo's guardrail layer itself changed; (3) the /grant operator channel (.claude/hooks),a wall-BYPASS mechanism, new attack surface; (4) the proxy-hardening campaign ledger + research (dev/, ~9.6k lines of data/docs). The body itself asks whether this should merge as one PR or be split.

Findings: 1 blocker, 1 high, 8 medium, 2 nit. (11 raw findings dissolved by evidence gates.)

Checked clean: mechanical.

Optional follow-up: draft ast-grep rules for recurring classes (F020, F021, F023) are available on request.

Additional observations:
- [2026-07-27] security: Bash tool writes to wall paths are not intercepted by the orchestrator (only Write/Edit/MultiEdit with a file_path are checked). This is a pre-existing bypass for .rules/, .claude/hooks/, etc., not introduced by this PR, but it means the grant state file is also forgeable via Bash echo even if it were added to WALL_PATHS. A complete fix requires provenance (signature/HMAC) or storing grant state outside the repo/filesystem accessible to the assistant.
- [2026-07-27] security: No mechanical scan hits landed on this shard; findings were derived from manual deep reading of the disclosed files and verification that .claude/state/walls-grant.json is outside WALL_PATHS.
- [2026-07-27] tests: Source files with added logic and no matching test change in this diff: .claude/hooks/modules/userpromptsubmit/03_grant_command.py (229 new lines, grant-channel security hook, zero test files anywhere), dev/campaigns/proxy-hardening/walls/*.py (campaign enforcement scripts, zero test files), checks/gate.sh (gate orchestrator additions, no dedicated test).
- [2026-07-27] tests: Rule-tests reviewed for the 3 new walls (kt-head-scoped-config-must-be-keyed-test, kt-shared-summary-parts-single-source-test, kt-no-runblocking-exported-bridge-test) and the heavily-modified walls (kt-no-silent-result-collapse-test, launcher-no-pkill-test, webui-fetch-only-in-api-tsx-test): each contains both valid and invalid cases, and spot-checks with ast-grep confirm the invalid cases trip their rules.
- [2026-07-27] tests: ExampleConfigTest parses the live config/splice.example.toml (walks up from the module directory), so it will fail if the committed example drifts; not a snapshot copy.
- [2026-07-27] tests: HeadServerReviewTest tear assertion at 24eefbf remains ordering-only and non-vacuous: message_start must not follow the error frame, without requiring presence that depends on where the premature EOF surfaces.
- [2026-07-27] build-config: Ledger status audit: all 88 items in dev/campaigns/proxy-hardening.toml have status = 'todo'; zero done/verified/in_flight/blocked items were found, so there is no status-inflation sample to audit.
- [2026-07-27] build-config: Registry reference audit: every non-empty wall= path in dev/campaigns/proxy-hardening/walls/wall_registry.toml (7 refs) and law_registry.toml (6 refs) resolves to an existing file in the repo checkout.
- [2026-07-27] build-config: Orchestrator dispatch audit: .claude/hooks/modules/userpromptsubmit/03_grant_command.py is the only module under that lifecycle, loaded alphabetically by the generic runner; no earlier module consumes .claude/state/walls-grant.json before the grant module writes it.
- [2026-07-27] build-config: Build/version audit: gateway/app/build.gradle.kts changes are limited to a tasks.test input declaration; package.json changes add script aliases only. No new dependencies or floating version ranges were introduced.
- [2026-07-27] correctness: kt-shared-summary-parts-single-source.yml: pattern 'summaryParts = $ANY' matches any assignment to a variable named summaryParts, not only TurnMeta.copy() re-assignment, so an unrelated local variable with that name would false-positive. Rule-test also uses property name 'summaryPartsShared' in valid/invalid cases, which never exercises the assignment branch (it only tests SharedSummaryParts() construction and the unrelated name).
- [2026-07-27] correctness: kt-l3-end-turn-literal.yml and kt-l3-sole-wire-terminals.yml: split-literal branches are limited to two-operand additive expressions, so three-way splits such as "end" + "_" + "turn" or "message" + "_" + "stop" bypass the rule. The PR comment explicitly documents the two-operand scope, so this is noted rather than raised as a finding.
- [2026-07-27] correctness: kt-head-scoped-config-must-be-keyed.yml: relational approach is mostly sound, but the 'val key = $ANY' arm could false-positive on an unrelated local named 'key' (e.g. a map key) in a function that also calls config.getConfig(). No such collision observed in current code.
- [2026-07-27] correctness: Turn.kt SharedSummaryParts default: exactly one construction site exists (TurnMeta default) and the new wall enforces no others. copy() preserves the reference, which is the intended turn-scoped sharing across continuation rounds. No drift observed from the round-1 verified state in Daemon.kt/ConfigService.kt/TurnPerf.kt/TurnDriver.kt/ResponsesStreamTranslator.kt/TurnTerminal.kt/UpstreamFailureClassifier.kt/ControlServer.kt.
- [2026-07-27] correctness: Mechanical scan hit at UpstreamFailureClassifier.kt:120 is a false positive; the when expression is on numeric ranges, not a sealed type. Round-1 adjudication holds.
- [2026-07-27] correctness: dev/walls-grant/03_grant_command.py and install.sh: grant lifecycle is operator-only by UserPromptSubmit construction, expiry is checked at the gate, and the install self-test verifies red/green/expired/module-issued paths. No correctness defect found.
- [2026-07-27] flowmap: SharedSummaryParts is updated optimistically before the wire write (ResponsesStreamTranslator.kt:508-512 records the part, then sink.thinkingDelta is called). In the abort-specific window the part may therefore be recorded but not yet flushed/acked by Ktor/Netty. Because an abort cancels the turn and no continuation round runs (fold/re-anchor gates on clientGone/watchdogFired), this does not currently suppress a part the next round's client never saw, but the seam between state and actual socket delivery is uninstrumented.
- [2026-07-27] flowmap: ControlServer has no explicit channel field; teardown is via EmbeddedServer.stop() and nulling the server reference (ControlServer.kt:143). No code in this file observes the Netty channel close, so the seed's claim that 'ControlServer's channel sees the teardown' does not map to an added field or callback in the changed code.
- [2026-07-27] RELIEF_AUDIT 2026-07-27 (round 2): scanned all 11 dissolutions; none re-examined. F001-F009 are round-1 recurrences dissolved on identical named evidence (immune memory stable). F011/F015 folded into F010 with facets preserved (Bash bypass, audit identity), consolidation not comfort-dissolution. The uncomfortable findings (forgeable grant state, no-op oracle check) SURVIVED with checkout-verified evidence.

.claude/hooks/orchestrator.py
marcospaulo: Closed in #59. Both halves landed; the first is stronger than prescribed.

Rather than adding the record to `WALL_PATHS` — which would leave it writable *under* an active grant, letting a grant extend itself — a tool-write to `.claude/state/walls-grant.json` is refused **unconditionally** (`NEVER_WRITABLE`, checked before the grant test). Records are HMAC-SHA256 signed with the signature verified **before** `until` is read; `session_id` and an optional `paths` scope are inside the signed bytes. The key lives at `~/.local/state/splice/walls-grant.key` (0600), minted only by `install.sh` — never by the gate, since a verifier that can mint its own key proves nothing. Verify/expiry/scope live once in `.claude/hooks/lib/walls_grant.py`, shared by the gate and `/grant` status so they cannot drift.

Your residual is recorded in that file's header rather than papered over: Bash never reaches PreToolUse, and a key readable by the process it defends is a bar, not a seal. The full seal (agent-inaccessible keyring, out-of-band issuer) is named as future work.

The forge is now a permanent red arm in the installer self-test — unsigned BLOCKED, hand-extended BLOCKED, expired BLOCKED, scope-mismatch BLOCKED, scope-match ALLOWED, module-issued E2E ALLOWED. It also proved itself live: the operator's own unsigned 90-minute grant stopped opening walls the moment the gate landed.

Two defects surfaced while proving it, neither in the review: `install.sh`'s idempotence check keyed on the *name* `_walls_grant_active`, so a repo carrying the pre-signature helper was reported "already grant-aware" and silently left forgeable; and the self-test's trailing `rm -f` silently revoked live grants mid-session. Both fixed.

dev/campaigns/proxy-hardening/oracle/capture.mjs
marcospaulo: Closed in #59. `--check` now compares three things, because each fails independently: every scenario fixture byte-for-byte, the vendored mock's sha256 (recomputed live, so nothing else saw its drift), and the scenario roster in **both** directions — a fixture the fresh run no longer produces is drift just as much as a new one.

Proved red by perturbing `server/src/anthropic/sse.mjs`: 6 fixtures drifted, exit 1. Green again after revert. `oracle:check` and INF-03's verify are real gates now.

dev/walls-grant/install.sh
marcospaulo: Closed in #59, taking the reconcile route rather than the recorded-hash one: a hash file in `.claude/state/` would be one more untracked file to trust, which is the same class of problem the grant record itself had. `install.sh` now diffs each installed menu file against its tracked canonical copy, reports drift loudly with the diff before overwriting, and asserts the copy landed. `/grant` status reports the same drift, since that is the screen an operator reads before trusting the channel.

I did not try the symlink route — you flagged it as unverified, and the reconcile is deterministic without depending on how Claude Code resolves command files.

gateway/app/src/test/kotlin/ExampleConfigTest.kt
marcospaulo: Closed on this PR (`d5ffbd8`), taking your preferred direction. Every camelCase comment token now resolves against `Knob.byKey`; prose lives in an explicit allowlist that is itself **asserted disjoint from the enum**, so an allowlist entry can never silence a real drift. Added a non-empty guard too, since a regex that stops matching would otherwise pass vacuously.

Confirmed your read before changing it: the example names only `firstByteTimeoutMs`, `maxInflight`, `maxQueued`, `streamIdleMs`, `upstreamTimeoutMs` and `xAI` today, so the old filter was lossless purely by accident. Proved red with a planted `toolSurfaceTypo`.

Follow-up: the widened assertion turned CI red on detekt (`ArgumentListWrapping`) — I had verified with `gradlew test`, and the gate runs `gradlew clean check`. Fixed in `3b66c10`.

.claude/hooks/modules/userpromptsubmit/03_grant_command.py
marcospaulo: Closed in #59. `/grant 30 --paths .rules <reason>` — a flag rather than a trailing positional, since a free-text reason and a bare path aren't distinguishable. Unknown paths are **refused, not dropped**: a typo'd path silently widening the grant to everything is the exact failure the feature exists to prevent. Omitting `--paths` keeps the previous all-paths behaviour so an existing habit doesn't silently become a refusal.

Sequenced after the forgeability fix as you advised, and `paths` sits inside the signed bytes — so a scope cannot be widened after issue without invalidating the signature. Both directions are red/green arms in the installer self-test.

.rules/kotlin-splice/kt-catch-swallows-cancellation.yml
marcospaulo: Closed in #59, exactly as directed. The exemption now binds `$V` and requires `throw $V` in the same catch block.

Confirmed both properties you predicted: the log-only body trips the wall, and rethrowing a *different* variable (`throw other`) also trips it — ast-grep does hold `$V` consistent across the `all`. Rule-test gained both as red cases. Verified 0 new hits across `gateway/*/src/main/**/*.kt`, so this is a pure widening with no fallout.

.rules/kotlin-splice/kt-no-silent-result-collapse.yml
marcospaulo: Partially closed in #59, and the other half is a deliberate refusal with evidence.

The `var` arm landed as you sequenced it — mechanical, red rule-test added, 0 new hits on the tree.

The return-position arm I validated and it **fails**. With the anchor relaxed to include `jump_expression`, this legitimately-handled shape trips the wall:

```kotlin
val r = runCatching { read() }
r.onFailure { log(it) }      // failure consumed, on its own statement
return r.getOrNull()
```

The `not: has onFailure` guards are scoped to the collapse node's subtree and the binding's subtree, and a standalone `onFailure` statement is in neither. Binding the receiver in a `not: follows: $RECV.onFailure($$$)` arm does not suppress it either (trailing-lambda call shape).

So it is left open and named in the rule header along with the failed experiment, rather than closed with a rule that cries wolf — your own argument from the `launcher-no-pkill` thread. Closing it properly needs failure-consumption tracking the pattern layer doesn't have.

.rules/rules/launcher-no-pkill.yml
marcospaulo: Closed in #59, and widened past the sketch. Your regex covers the `pk|ill` split only, but `'p' + 'kill'` produces the identical runtime command and would still have walked through. The branch now requires the morpheme at **all four** split points:

```
(p['"`]\s*\+\s*['"`]kill|pk['"`]\s*\+\s*['"`]ill|pki['"`]\s*\+\s*['"`]ll|pkil['"`]\s*\+\s*['"`]l)
```

Backticks included, so the template-literal edge you flagged is covered rather than deferred. Rule-test pins all four red plus `'ls' + ' -la'` and `'echo ' + 'hello'` green — your point about a wall that flags compliant code teaching people to reach for `ast-grep-ignore` is exactly why the green cases are pinned too. 0 hits across `server/**/*.mjs`.

dev/campaigns/proxy-hardening.toml
marcospaulo: Closed in #59, with one deliberate difference. I did **not** point the default at `proxy-hardening.toml`: seven ledgers live in `dev/campaigns/`, and silently guessing one for a mutating verb (`note`, `set-status`, `claim`) would write the right text into the wrong campaign — strictly worse than the traceback it replaces.

Instead a path-less verb now names the problem and prints the ledger census, so the tool teaches instead of requiring the NOTE to have been read. `laws` still works with no path, since it legitimately aggregates.

The header NOTE then shrank to a pointer as you suggested — done through `manifest.py amend-header`, which already existed but was missing from USAGE; it's documented now, since it is the sanctioned channel for exactly this and hook 08 blocks the raw edit.

gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
marcospaulo: Closed on this PR (`d5ffbd8`) — I took the encapsulation rather than leaving the KDoc, since "the discipline is a property of the type, not the comment" is this repo's own doctrine. Collections are private behind `partAt` / `markEmitted`; I dropped your `emittedCount` because nothing reads it and the recap arm's `cursor in indices` check folds into `partAt(cursor) == part` (`RECAP_DONE = -1` still yields null, so the semantics are unchanged).

Your finding forced a bigger one into the open, which I'd have missed otherwise: `StreamTurnContext.summaryPartsShared` was `SharedSummaryParts? = null` for test convenience, and a null there silently restores the per-round private dedup state this whole PR exists to remove. It is now **required with no default** — production passes `meta.summaryParts`, and a test that wants round-private state has to say so. Four test construction sites updated.

.claude/hooks/modules/userpromptsubmit/03_grant_command.py
marcospaulo: Closed in #59 with your second resolution — the tolerant shape kept, with a comment recording the trace so the next reader doesn't repeat it.

I could not confirm from this repo which payload shapes the CLI has shipped, and guessing wrong disables the operator's only legitimate channel, so tightening to `!= "UserPromptSubmit"` wasn't a safe trade. The comment now names the three layers that make the tolerance affordable: placement under `modules/userpromptsubmit/`, the `prompt`-is-a-string gate two lines below, and — new since round 2 — the fact that a hand-rolled invocation still cannot produce a grant the gate will honour without the signing key. That last one is what actually retires your "dominated by the state-file issue" caveat.

dev/campaigns/proxy-hardening/walls/wall_registry.toml
marcospaulo: Closed in #59, taking your second option — the one that cannot rot. The numbers are gone rather than corrected:

> THE COUNT IS THE ROWS. Do not trust any summary, INCLUDING THIS ONE.

with a pointer to `npm run gate:campaign:census`, which computes `items N | walled N | UNWALLED N | ...` from the rows. Your count was right (88 / 81 / 7) and the gate prints exactly that, which is precisely why the header shouldn't be the one repeating it.

Getting here needed a guard fix first: hook 08 blocked `wall_registry.toml` as a "campaign ledger", but it isn't one — no `[[items]]`, and `manifest.py` has no verb that can touch it. It offered a CLI alternative that doesn't exist, i.e. a wall with no door. Narrowed to direct `dev/campaigns/` children, red/green proved both ways.

PR #59, opened 2026-07-27
fix(walls): sign the grant record — it was forgeable in one Write

marcospaulo:
Extracted from #58, which the round-2 review flagged as carrying four unrelated payloads. This is
the self-contained one: the guardrail layer and its operator channel. **No Kotlin source** — so
`check`/detekt on this branch is identical to `main`.

## Why this is its own PR

The forgeable-grant hole below survived a full review round because it landed *inside* a PR about
streaming. A security change reviewed as an appendix to a streaming fix is how that happens.

## 1. The blocker — `/grant` was forgeable in one Write

The channel was well engineered on every path it controlled, and did not control the one thing the
gate trusts: the FILE. `.claude/state/` was not a wall path, so an assistant could Write its own
`walls-grant.json`; `_walls_grant_active()` trusted `until` with no provenance check; `.gitignore`
kept the forgery out of git. "Operator-only by construction" was true of ISSUING and false of
TRUSTING — and only the second one matters to a gate.

Grants are now HMAC-SHA256 signed, verified **before** `until` is read, key at
`~/.local/state/splice/walls-grant.key` (0600) minted only by `install.sh` — never by the gate,
since a verifier that can mint its own key proves nothing. Verify/expiry/scope live once in
`.claude/hooks/lib/walls_grant.py`, shared by the gate and `/grant` status so they cannot drift.
The record is `NEVER_WRITABLE`: refused as a tool-write target unconditionally, not merely walled,
because a grant extending itself is never legitimate. Every failure shape returns "no grant".

**What this does not close is in that file's header rather than papered over:** Bash never reaches
PreToolUse, and a key readable by the process it defends is a bar, not a seal.

Grants also scope now — `/grant 30 --paths .rules <reason>`, inside the signed bytes, unknown paths
refused rather than dropped — and `session_id` is signed, so the audit record answers who and from
where.

Self-test matrix, all arms permanent: unsigned BLOCKED, hand-extended BLOCKED, expired BLOCKED,
scope-mismatch BLOCKED, scope-match ALLOWED, module-issued end-to-end ALLOWED.

### Two defects found while proving it, neither in the review

- `install.sh`'s idempotence check keyed on the *name* `_walls_grant_active`, so a repo carrying the
  pre-signature helper was reported "already grant-aware" and silently left forgeable — an installer
  skipping the upgrade it was run to perform. Now keys on the signed gate and fails loudly.
- The self-test's trailing `rm -f` silently **revoked any live grant**, so re-running the installer
  mid-session made an operator's wall writes start failing with no explanation. Byte-exact trap
  restore, safe only because grants are signed.

## 2. The frozen oracle's `--check` was vacuous

It re-captured to a temp dir, deleted it, and exited on scenario success alone — it never read the
committed fixtures, so no behaviour drift could trip it, and it is wired as `oracle:check` and as
ledger item INF-03's verify. Now diffs fixture bytes, the vendored mock's sha, and the scenario
roster both directions. Proved red by perturbing `server/`, green after revert.

## 3. The ast-grep walls audit (33 -> 36 walls, 29 rule improvements)

Nearly every wall was narrower than its own message claimed. Round 2 found three more:

- `kt-catch-swallows-cancellation` accepted a type-check without the rethrow its own message
  demands. Now binds `$V` and requires `throw $V` — rethrowing a *different* variable still trips.
- `launcher-no-pkill`'s split-literal branch fired on any two-literal concat in exec arguments
  (`execSync('ls' + ' -la')` tripped a pkill wall). Now requires the morpheme at **all four** split
  points, not just the `pk|ill` shape.
- `kt-no-silent-result-collapse` gains the `var` binding. The return-position arm was validated,
  **refuted** (it false-positives on a Result handled by a standalone `onFailure` statement) and
  left open with the failed experiment recorded in the header — a wall that cries wolf teaches
  people to bypass every wall.

## 4. Campaign ledger + registry

Hook 08 blocked `wall_registry.toml` — not a ledger, no CLI verb, a wall with no door. Narrowed to
direct `dev/campaigns/` children. The registry header then dropped its hand-transcribed census for
a computed one. `manifest.py`'s default named a ledger vendored from another repo, so the lawful
path-less invocation tracebacked; it now names the problem and prints the census, deliberately
without guessing among seven ledgers.

## Gates

`scan` clean, rules 34/34, hooks 13/13, campaign selftest, manifest selftest, oracle check clean,
server 104/104, installer self-test 8/8 arms — all verified with this branch standing alone on
`main`.

## Merge order

Merge this first, then rebase #58: these commits drop as patch-equivalent (`03867d2` here and on
#58 share patch-id `603acff4`), and #58 slims back to its titled subject.


marcospaulo:
Closing: superseded by #58's squash-merge (\`561d7ec\`), which absorbed this branch's entire payload.

Verified before closing — all **102 files** this branch touches are byte-identical on \`main\`. This PR contributes nothing, and merging it would be actively harmful: the diff from \`main\` to this head is **+69 / −933 across 32 files**, every one a reversion (it would delete \`DaemonPerHeadConfigTest\`, \`SummaryTurnDedupTest\`, \`HeadServerReviewTest\`, both per-head config walls, the early-open handoff, the \`SharedSummaryParts\` encapsulation and the \`ExampleConfigTest\` widening).

Why the split failed: the walls commit was pushed to #58 as well, so the split only held if this PR merged first. #58 merged first and was **squash**-merged, and a squash has no patch-id matching any individual commit — so the "commits drop as patch-equivalent on rebase" property this split relied on never applied.

Nothing is lost. Everything here is on \`main\` and green, including the grant-forgery fix. Pre-split history is preserved at \`backup/pr58-preseplit-20260727\` if it is ever wanted.

PR #62, opened 2026-07-28
fix(walls): close the two walls the 2026-07-26 audit deferred

marcospaulo:
The audit left `kt-no-println` and `kt-no-system-getenv` unlanded pending an operator call
(§2.1/§2.2). Both are settled — and they resolved in **opposite** directions.

## `kt-no-println` — widened, and the 13 sites converted

`pattern: println($$$A)` matches a bare call only, so the shape that actually occurs here was
invisible: `System.err.println` is a navigation_expression, `System.err::println` a different one,
`::println` a callable_reference. **The wall reported 0 while 13 live sites bypassed the endpoint.**

The invariant was *verified* before widening, not assumed. `Main.persistentLogger` writes **both**
stderr and `daemon.log`; `/mgmt/logs` serves `daemon.log`. A bare `System.err.println` reaches
stderr only — so auth failures, config-persist failures and the tool-surface latch signal were all
genuinely unreadable through the API.

All 13 converted, not exempted, via an injected `log: (String) -> Unit`.

**One carve-out, reasoned inline:** `AsyncFileIo.recordDrop` keeps stderr, because
`persistentLogger` writes `daemon.log` *by calling* `AsyncFileIo.submit`. Routing its drop-warning
through the sink would re-enter the component that just failed — the one message you most need
becomes the one guaranteed not to arrive.

## `kt-no-system-getenv` — widened, but not as drafted

The audit proposed `callable_reference` branches, surfacing 21 hits. I read all 21: **every one is
injection, not access** — a DI default (`env: (String) -> String? = System::getenv`) or the reader
handed to a function that asked for it. That is the identical idiom `ConfigService.kt:37` is
exempted for. Landing it would flag this codebase's own constructor-injection convention and turn
CI red for a rule nobody agreed to.

The real defect is narrower: `java.lang.System.getenv("HOME")` is a different node and walked
straight through. `$RECEIVER.getenv($$$A)` closes it and scans to **0 hits**.

> A CALL reads the environment; a REFERENCE hands the capability to someone who asked for it.
> Only the first bypasses the config layer.

## Why `DaemonLog` rather than 12 lines in `Daemon`

Threading the sink explicitly through twelve construction sites pushed `Daemon` past detekt's
`LargeClass` ceiling — it sits at 524 lines against a 400 threshold, so *any* addition tips it.
I did not weaken the gate, and I did not force a `Daemon` split as a side effect of a logging fix.
The sink is a process default installed once by `Main`; injection remains the contract, and
uninstalled it is a **no-op, never a silent stderr write**. `Daemon` is byte-identical to `main`.

## One test caught the change, and the test was fixed — not the code

`CodexProviderTest` asserted the latch line on captured **stderr**. Post-change that would pass
*vacuously with zero lines* — exactly the regression the assertion exists to catch. It now captures
the injected sink; its orphaned `captureStderr` helper and two imports are gone.

## Proof

- widened `kt-no-println` on unmodified code → **13 hits**; after conversion → **0**; reverting a
  single site re-trips it
- `kt-no-system-getenv` → **0 hits**, verified against the tree and a synthetic file carrying each shape
- rule-tests **2 → 7 cases each**, pinning both directions
- `npm run gate`: **PASS**

### DIFF
--- a/.rules/kotlin-splice/kt-no-system-getenv.yml
+++ b/.rules/kotlin-splice/kt-no-system-getenv.yml
@@ -1,6 +1,25 @@
 # Config discipline — env access goes through the layered config service (the
 # Kotlin analog of webui's fetch-only-in-api). Scattered System.getenv reads
 # bypass provenance, coercion, and the /mgmt/config layer view.
+#
+# WIDENED 2026-07-27 (audit §2.1) — but NOT the way that audit drafted it, and the difference is
+# the whole point. The audit proposed adding `$RECEIVER::getenv` and a bare `System::getenv`
+# callable_reference branch, which surfaces 21 live hits across 11 files and calls them violations.
+# Every one of those 21 was read: all are INJECTION, not scattered access — either a DI default
+# (`env: (String) -> String? = System::getenv`, e.g. TopologyLoader.configPath, InstallCommand,
+# AdminSupport, DoctorCommand) or the reader handed to a function that takes it as a parameter
+# (`GrokOAuthEndpoints.tokenUrl(System::getenv)`). That is the IDENTICAL idiom ConfigService.kt:37
+# uses and is exempted for. Banning it would flag this codebase's own constructor-injection
+# convention and turn CI red for a rule nobody agreed to.
+#
+# The real defect is narrower: the rule saw only `System.getenv(...)`, so the fully-qualified
+# spelling `java.lang.System.getenv("HOME")` — a different node (navigation_expression with a
+# qualified receiver) — walked straight through. That is a genuine evasion of a real read.
+# Adding `$RECEIVER.getenv($$$A)` closes it, leaves all three injection forms clean (proved
+# against the tree AND against a synthetic file carrying each shape), and scans to 0 hits.
+#
+# The distinction this rule now draws: a CALL reads the environment; a REFERENCE hands the
+# capability to someone who was asked to accept it. Only the first bypasses the config layer.
 id: kt-no-system-getenv
 language: kotlin
 severity: error
@@ -10,4 +29,8 @@ files:
 ignores:
   - gateway/core/src/main/kotlin/**/config/**
 rule:
-  pattern: System.getenv($$$A)
+  any:
+    - pattern: System.getenv($$$A)
+    # fully-qualified spelling — java.lang.System.getenv("X"); a different node the bare
+    # pattern cannot reach, and the only real evasion the old rule had.
+    - pattern: $RECEIVER.getenv($$$A)
--- a/.rules/kotlin-splice/kt-no-system-getenv.yml
+++ b/.rules/kotlin-splice/kt-no-system-getenv.yml
@@ -12,11 +12,16 @@
 # uses and is exempted for. Banning it would flag this codebase's own constructor-injection
 # convention and turn CI red for a rule nobody agreed to.
 #
-# The real defect is narrower: the rule saw only `System.getenv(...)`, so the fully-qualified
-# spelling `java.lang.System.getenv("HOME")` — a different node (navigation_expression with a
-# qualified receiver) — walked straight through. That is a genuine evasion of a real read.
-# Adding `$RECEIVER.getenv($$$A)` closes it, leaves all three injection forms clean (proved
-# against the tree AND against a synthetic file carrying each shape), and scans to 0 hits.
+# The real defect is narrower: the rule saw only `System.getenv(...)`, so other spellings of a real
+# read walked straight through — `java.lang.System.getenv("HOME")` (navigation_expression with a
+# qualified receiver) and, per the #62 review, a static import (`import java.lang.System.getenv`
+# then a bare `getenv("HOME")`, which is a call_expression with a plain identifier callee and
+# matches NEITHER of the receiver-shaped branches). Three branches now cover all three spellings.
+#
+# An earlier revision of this header called the qualified receiver "the only real evasion". That
+# was wrong and the static-import case proves it: this rule matches SPELLINGS, and an enumeration
+# of spellings is never provably complete. Treat the branch list as the shapes known to occur, not
+# as closure — if you find a fourth, add it rather than trusting this comment.
 #
 # The distinction this rule now draws: a CALL reads the environment; a REFERENCE hands the
 # capability to someone who was asked to accept it. Only the first bypasses the config layer.
@@ -32,5 +37,11 @@ rule:
   any:
     - pattern: System.getenv($$$A)
     # fully-qualified spelling — java.lang.System.getenv("X"); a different node the bare
-    # pattern cannot reach, and the only real evasion the old rule had.
+    # pattern cannot reach.
     - pattern: $RECEIVER.getenv($$$A)
+    # static import — `import java.lang.System.getenv` then a bare `getenv("X")`. No receiver at
+    # all, so neither branch above can see it. `kind: call_expression` is load-bearing: without it
+    # the pattern also binds the `getenv` identifier inside the two receiver forms above.
+    - all:
+        - pattern: getenv($$$A)
+        - kind: call_expression
--- a/gateway/app/src/main/kotlin/splice/app/Main.kt
+++ b/gateway/app/src/main/kotlin/splice/app/Main.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeoutOrNull
 import splice.core.config.StatePaths
 import splice.core.util.AsyncFileIo
+import splice.core.util.DaemonLog
 import splice.core.util.runCatchingCancellable
 import java.nio.file.Files
 import java.nio.file.Path
@@ -48,6 +49,10 @@ private fun runDaemon() {
     val topology = TopologyLoader.loadOrMaterialize(topologyPath)
     val distPath = Paths.get(System.getProperty("user.dir"), "..", "webui", "dist", "index.html")
     val log = persistentLogger(statePaths.logsDir)
+    // Components that would otherwise fall back to bare stderr (auth providers, ConfigService,
+    // ResponsesProvider) default to this sink, so their diagnostics reach daemon.log and therefore
+    // /mgmt/logs. Injection still wins where a caller passes its own (wall kt-no-println).
+    DaemonLog.install(log)
     val shutdownSignal = CompletableDeferred<Unit>()
     splice.app.cli.shimStalenessWarning()?.let { log("$it\n") }
     val daemon = Daemon(
--- a/gateway/app/src/main/kotlin/splice/app/Main.kt
+++ b/gateway/app/src/main/kotlin/splice/app/Main.kt
@@ -123,14 +123,23 @@ private const val TEARDOWN_TAIL_GRACE_MS = 2_000L
 // stays OPEN for the daemon's lifetime. A failed write drops the writer so the
 // next line reopens. SIZE-ROTATION (audit 2026-07-18: daemon.log grew forever, ~1KB/turn): at
 // MAX_LOG_BYTES the file is moved to daemon.log.1 (one generation kept) and a fresh file opened.
-private fun persistentLogger(logsDir: Path): (String) -> Unit {
+//
+// LINE TERMINATION IS NORMALIZED HERE, not trusted to callers (review of PR #62, 2026-07-27).
+// This sink used to `print`/`write` the message verbatim, so a trailing "\n" was part of every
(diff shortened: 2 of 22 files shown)


gateway/app/src/main/kotlin/splice/app/Main.kt
```diff
@@ -48,6 +49,10 @@ private fun runDaemon() {
     val topology = TopologyLoader.loadOrMaterialize(topologyPath)
     val distPath = Paths.get(System.getProperty("user.dir"), "..", "webui", "dist", "index.html")
     val log = persistentLogger(statePaths.logsDir)
+    // Components that would otherwise fall back to bare stderr (auth providers, ConfigService,
+    // ResponsesProvider) default to this sink, so their diagnostics reach daemon.log and therefore
+    // /mgmt/logs. Injection still wins where a caller passes its own (wall kt-no-println).
+    DaemonLog.install(log)
```
marcospaulo, line 55: **[blocker]** The persistentLogger design itself is solid (bounded AsyncFileIo lane, size rotation, drop-safe writer reopen), and the five pre-existing callers follow its convention of supplying their own newline (Main.kt:57, AuthProbeLoop.kt:58/75/76, Daemon.kt:133). The converted sites do not, and the sink does not add one either: persistentLogger emits `System.err.print(line)` and `w.write(line)` with no terminator (Main.kt:134-146). I checked the converted call sites (RefreshOutcome.kt:56-72, ConfigService.kt:175,194, CodexAuthProvider.kt:128,329, GrokAuthProvider.kt:144,271,346, KimiAuthProvider.kt:209,264, ApiKeyAuthProvider.kt:81) and none of the messages contains a newline.

The failure is reproducible: trigger any two converted log events (for example two "refresh skipped" lines) and tail daemon.log; both entries land on a single line, and /mgmt/logs serves exactly that file (ControlServer.logsJson -> LogFileSource.tail). Since routing these diagnostics to /mgmt/logs is the point of this PR, this one seems worth fixing before merge.

Direction: terminate in the sink, one place, and drop the explicit `\n` from the five legacy call sites so the terminator becomes sink-owned:

```kotlin
return { msg ->
    val line = "[${LocalTime.now().truncatedTo(ChronoUnit.SECONDS)}] $msg\n"
    AsyncFileIo.submit {
        System.err.print(line)
        ...
        w.write(line)
    }
}
```

The alternative (appending `\n` at each of the 14 converted sites) also works, but it keeps termination as a per-caller obligation, which is how this slipped in. Sink-side is my recommendation rather than a requirement; what matters is that exactly one side owns the newline.

.rules/kotlin-splice/kt-no-system-getenv.yml
```diff
@@ -10,4 +29,8 @@ files:
 ignores:
   - gateway/core/src/main/kotlin/**/config/**
 rule:
-  pattern: System.getenv($$$A)
+  any:
```
marcospaulo, line 37: **[suggestion]** The widening does what it claims for the shapes it targets. I verified against the shipped rule: `System.getenv("HOME")` and `java.lang.System.getenv("HOME")` both match, and the `System::getenv` injection idiom stays clean, so the call-vs-reference distinction in the rule comment is working as designed.

One spelling of the same evasion class is still open: a static import. `import java.lang.System.getenv` followed by a bare `getenv("HOME")` matches neither branch, because both require a receiver. I confirmed this by running this PR's rule against a synthetic file (zero matches on the bypass shape). The rule comment says the fully-qualified spelling was "the only real evasion the old rule had"; the static-import spelling is the same class (a real env read through an alternate node shape), so the wall is currently one spelling short of the closed status it documents.

On impact, being precise: there is no violation in the tree today (the rule scans to 0 hits), and this shape takes a deliberate, unusual import that review would likely catch. That is why this is a suggestion, not a concern. Given the walls are the drift-control mechanism here, closing it is one branch:

```yaml
- pattern: getenv($$$A)
  inside:
    stopBy: end
    has:
      stopBy: end
      pattern: import java.lang.System.getenv
```

Gating the bare call on the import keeps a locally-defined `fun getenv(...)` clean (no import, no match), while the receiver forms stay with the existing branches. I tested this branch against the violation and the compliant shapes and it behaves as described. An alternative is prohibiting the static import itself; I mildly prefer matching the call, since the call is the actual read, but either closes the hole.

gateway/core/src/main/kotlin/splice/core/util/DaemonLog.kt
```diff
@@ -0,0 +1,36 @@
...
+// Install-once at process start, read-only after. Uninstalled (tests, library use) it is a NO-OP
+// rather than stderr: a component that wants output says so by injecting a sink.
+package splice.core.util
+
+public object DaemonLog {
+    @Volatile
+    private var sink: (String) -> Unit = {}
+
+    /** Called once from Main with the persistent logger (writes both stderr and daemon.log). */
+    public fun install(logger: (String) -> Unit) {
```
marcospaulo, line 28: **[suggestion]** DaemonLog's design reads well: the header explains why a process default exists at all (threading a sink through twelve construction sites pushed Daemon past the LargeClass ceiling), the uninstalled default is a no-op rather than a silent stderr write, and CodexProviderTest injecting a capturing sink is exactly the right consumer pattern.

The gap is that DaemonLog itself is exercised nowhere. No test calls install, asserts write forwards to the installed sink, or pins the uninstalled no-op contract (grep across gateway/*/src/test: zero DaemonLog references). Every converted-site test injects its own sink, so the default path production actually uses (Main.kt:55 install, then DaemonLog::write defaults in nine providers) has no coverage. The falsifiable consequence: if `write` stopped invoking `sink`, or Main dropped the install call, all twelve routed diagnostics would silently vanish and the whole suite would still pass.

One companion observation, same file: the object has install + write and no reset (DaemonLog.kt:28-35). The first test that does exercise install will leak its sink into the rest of the JVM run unless a reset exists, so the natural change is a small test plus a reset hook called from teardown:

```kotlin
class DaemonLogTest {
    @AfterEach
    fun reset() = DaemonLog.resetForTests()

    @Test
    fun `uninstalled sink is a no-op`() {
        DaemonLog.write("nothing should happen")
    }

    @Test
    fun `write forwards to the installed sink`() {
        val seen = mutableListOf<String>()
        DaemonLog.install(seen::add)
        DaemonLog.write("hello")
        assertEquals(listOf("hello"), seen)
    }
}
```

The coverage ask is the part I would treat as a requirement; the reset API shape is a preference (a guard that throws on a second install is the alternative, but it conflicts with legitimate test use, which is why I lean toward reset). I might be missing indirect coverage elsewhere; I grepped for DaemonLog under src/test and found nothing.

marcospaulo (commented):
## Review: fix(walls): close the two walls the 2026-07-26 audit deferred

**Intent:** widen `kt-no-println` / `kt-no-system-getenv` to the node shapes that were bypassing them, convert the 13 newly-visible `System.err.println` sites to an injected sink routed to `/mgmt/logs` via the new `DaemonLog` process default, and keep getenv callable-references exempt as DI idiom.

**Pipeline:** ast-grep mechanical scan (98 rules: bundled + this repo's own `kotlin-splice` and `kotlin` packs) plus 4 review lenses and a flow-map of the DaemonLog install/emit seam; every finding gated on quoted evidence. 25 raw findings in, **3 survived** (1 blocker, 2 medium), 22 dissolved with named gates. The dissolves were mostly the generic bundled pack (`no-lambda-seam`, `no-object-singleton-state`) firing on this repo's sanctioned constructor-injection idiom, plus api-abi constructor-break claims that have no victim (gateway modules are unpublished, single-build; verified in the gradle config), the test `runBlocking` hits (the repo's own walls cover `src/main` only, and `runTest` cannot virtualize a `Dispatchers.Default` SupervisorJob drain), and the raw-env-value logging concern (the ~30 knobs in `Knob.kt` are non-credential and that line fires only on non-numeric `NUMBER` values).

**Checked clean:** latch assertion in `CodexProviderTest` is non-vacuous (fails when zero lines arrive at the sink); rule tests pin both directions per widened branch (7 cases each, matching the description); the `AsyncFileIo.recordDrop` stderr carve-out is justified (the re-entrancy cycle through `AsyncFileIo.submit` is real); every converted site's default sink resolves to `DaemonLog::write` (no divergent defaults); `Daemon.kt` untouched as claimed.

**One blocker:** the installed sink concatenates lines. The 14 converted call sites pass messages with no trailing newline (verified: none contain `\n`; the 5 pre-existing `persistentLogger` callers supply their own), and `persistentLogger` writes them verbatim, so `daemon.log` and the `/mgmt/logs` output this PR exists to fix come out merged.

**Optional follow-up (not a finding):** the `kt-no-system-getenv` comment claims the fully-qualified spelling was "the only real evasion", but a static import (`import java.lang.System.getenv` + bare `getenv("X")`) matches neither branch (verified with ast-grep). A validated rule branch closing it is included in the inline comment if you want it.

gateway/app/src/main/kotlin/splice/app/Main.kt
marcospaulo: Fixed in `f9b3f60` — confirmed exactly as reported, and it was a real regression.

The sink printed/wrote the message verbatim, so a trailing `\n` was part of every caller's string **by convention**. That convention is invisible at the call site and silently breaks whoever forgets — and the kt-no-println conversion moved 14 sites off `System.err.println`, which appends the terminator for you. Their entries merged into one run-on line, and `/mgmt/logs` splits on `\n`, so the endpoint this whole change exists to feed emitted concatenated garbage.

Fixed at the **sink**, not across the 14 call sites: `"[...] ${msg.trimEnd('\n')}\n"` appends exactly one terminator, a no-op for the five legacy callers that already pass one. Patching the callers would have left the invisible convention armed for the next person; normalizing here makes the class unrepeatable.

`private` -> `internal` solely so the sink is testable — see `DaemonLogWiringTest`, which pins this: three messages must produce three lines. Proved non-vacuous by reverting the fix (the test fails; the messages merge) and restoring it.

.rules/kotlin-splice/kt-no-system-getenv.yml
marcospaulo: Fixed in `f9b3f60`, and you were right that the comment overstated closure — I corrected the claim rather than extending it.

Third branch landed for the static import. I took `all: [pattern: getenv($$$A), kind: call_expression]` rather than the import-gated form, and your read of the difference is exactly right: mine matches *any* bare `getenv(...)` call, so a locally-defined `fun getenv` would trip it. Worth being precise about my own reasoning there — that breadth was **not** a deliberate trade-off. I validated that `config.getEffort(...)` stays clean but never tested a locally-defined `getenv`. It happens to be acceptable (fails closed, 0 hits today, no `inside`/`has` machinery), but "acceptable in hindsight" and "chosen" are different things.

`kind: call_expression` is load-bearing, as you noted: without it the pattern also binds the `getenv` identifier inside the two receiver-shaped forms, which would flag the sanctioned `System::getenv` injection idiom.

The header no longer claims the qualified receiver was "the only real evasion". It now says the rule matches **spellings**, that an enumeration of spellings is never provably complete, and that the branch list is the shapes known to occur — so the next person adds a branch instead of trusting the comment. Rule-test 7 -> 8 cases.

gateway/core/src/main/kotlin/splice/core/util/DaemonLog.kt
marcospaulo: Fixed in `f9b3f60` — and this was the sharpest of the three, because it explains the blocker.

The production path had zero coverage, and the gap was not theoretical: **the newline regression shipped through a fully green suite**, precisely because every other test injects its own sink and the default is never exercised.

`DaemonLogWiringTest` covers both halves:
- output shape — three messages produce three lines, and a caller that already terminates is not double-terminated
- the wiring — `Main` install -> `DaemonLog::write` -> the provider default, including the one-hop binding test that catches a re-pointed default

On the reset hook: I used `DaemonLog.install {}` in a `finally` rather than adding a `resetForTests()`. Same no-op restoration with zero API growth — a test-only method on a production object is surface I would rather not add for something a `finally` already does. The sink is process-wide, so a leaked capture would contaminate every later test in the JVM.

Follow-up you and I both flagged, now also in this PR (`123b8cf`): the installer's self-test drove the *gate* end to end but never executed the `/grant` bang-line — the operator's only status surface — which is the same tested-path/used-path drift one level up. It now executes that line across four arms, including a reason carrying `( ) ; &` and a backtick. Proved non-vacuous, and my first attempt at that proof **was** vacuous: reverting only the installed copy passed, because the installer reconciles it from canonical before the test runs.

PR #68, opened 2026-07-29
fix(build): read detekt and junit versions from the catalog, and correct the squash claim

marcospaulo:
Three items, all fallout from the last round.

## 1. Version-catalog skew — the inventory's §4.2, plus one found beside it

`splice.kotlin-common.gradle.kts` pinned `detekt-formatting:1.23.8` **and** `junit-bom:6.1.2` as bare
literals while `detekt` and `junit` already lived in `libs.versions.toml`. A catalog bump left both
behind silently. Nothing fails loudly — you get skew, which is precisely what a version catalog
exists to make impossible.

A precompiled script plugin gets no generated `libs` accessor, which is why the literals were there;
`VersionCatalogsExtension` is the supported route from that context.

- resolved `detekt-formatting` → **1.23.8**, matching the catalog
- resolved `junit-bom` → **6.1.2**, matching the catalog
- a missing alias **throws** rather than falling back — proved red by aliasing `detekt-typo` and
  getting `version catalog has no `detekt-typo``

## 2. The claim I shipped in #67 is wrong, and this corrects it the same day

#67 says the PR title "replaces your branch commits in `main`". **It does not always.**

**#66 passed `lint` with the title `chore(reasoning-cache): …` and landed on `main` as
`verify(reasoning-cache): …`** — its *branch commit* subject, a type the linter forbids. So a green
`lint` does **not** guarantee `main`'s history complies, which is the opposite of the reassurance #67
gave.

Mechanism undetermined, and said so rather than guessed: it is not a commit-count rule (#66 had 3
commits and used the commit message; #65 had 1 and used the title).

The practical rule costs nothing and doesn't depend on knowing why: **use an allowed type in the
branch commit subject too**, and `main` is compliant whichever source the squash picks. `AGENTS.md`
and `CONTRIBUTING.md` now say that instead of the false reassurance.

`main` already carries one non-compliant subject (`verify(...)` from #66). Left as-is — rewriting a
merged `main` to fix a subject line isn't worth it, and it stands as the evidence for why the
corrected rule exists.

## 3. Inventory refreshed

- §1 `reasoning-cache` → closed by #66
- §4.3 14 stale worktrees → **removed**, re-verified inert immediately before (0 dirty, 0 open
  handles, both commits ancestors of `main`); **81 MB** recovered. One unrelated worktree under
  another session's scratchpad left alone.

`npm run gate`: PASS.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,10 +146,19 @@ check. Allowed types, and this is the complete list:
 Scope optional (`fix(walls): ...`). Anything else fails `lint` and blocks the merge.
 
 WHY THIS IS EASY TO GET WRONG THREE TIMES IN A ROW (it happened, 2026-07-28/29): branch commit
-messages are NOT linted, so `harden(walls):` and `verify(x):` write and review fine locally. The
-title is checked only after the PR exists — and because merges are SQUASHED, the PR title replaces
-your branch commits in `main`, so history shows only compliant types and gives no hint that a wider
-vocabulary was ever tried and rejected. The feedback arrives late and then erases its own evidence.
+messages are NOT linted, so `harden(walls):` and `verify(x):` write and review fine locally, and the
+title is checked only after the PR exists — late, well after the work reads finished.
+
+AND THE CHECK IS NOT SUFFICIENT. Corrected 2026-07-29, same day the original claim was written: this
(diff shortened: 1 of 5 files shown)


marcospaulo:
@torad-review review

PR #94, opened 2026-08-11
feat(gateway)!: delete the legacy Node stack, close the W4 hardening wave

marcospaulo:
34 commits. Two things dominate: the W4 correctness/honesty wave, and the removal of the legacy Node stack now that Kotlin has been serving production alone.

## BREAKING — `server/` is gone

The Kotlin daemon has owned `:3099` and `:3096` since 2026-08-07 23:48 and served **32,326 turns at 99.14% clean** with zero watchdog fires; no Node process has run in three days. `checks/cutover_ledger_gate.py` reported READY (37 verified, 0 unverified) before anything was removed.

Deleted: the `server/` tree, both Node shims, the six JavaScript ast-grep rules scoped to `server/**` **and their six rule-tests** (deleting a rule without its test breaks `gate:rules`), and one experiment script that drove its A/B by spawning the Node proxy. `bin/claudex-kt` took the name `bin/claudex`. Rewired: workspaces, lockfile, `checks/gate.sh`, README/AGENTS.md, `.rules/README`. CI needed no change — it already sets up JDK 21 and runs `npm run gate`.

Three things that would have broken silently and did not:
- the **migration oracle** read `server/test/codex-proxy.test.mjs` at runtime; the mock is now vendored, re-verified 11/11 *with `server/` deleted*
- `oracle/capture.mjs` can never run again; it exits 2 explaining why and is **kept** as the fixtures' provenance
- the **hook tests** fixtured entirely on the deleted JS rules — 4 of 12 went red on deletion, re-based onto `kt-l3-sole-wire-terminals` / `kt-embedded-server-loopback`

## W4 hardening

**CX-02** the compaction directive reaches all three dialects from one `:core` definition (Responses bytes sha-identical). **CX-09** the empty-turn honesty gate reads a new `emittedThinking` signal. **CX-11** `is_error` is parsed and the loop guard prefers it over an uncanaried string. **CX-18** one shared usage alias chain, adopted by all four readers.

**Adversarial review refuted the first cut of two of these before they shipped** — CX-09 as first written converted every thinking-only kimi turn into an `API_ERROR` after its content had already streamed, and CX-18's "one shared chain" claim was false (it added a fourth reader). Both were rebuilt; all four campaign walls were rewritten after reviewers proved five bypasses, and every bypass is now RED.

## Verification

`npm run gate` PASS · migration oracle **11/11 byte-exact** · campaign wall gate PASS (earned-green 25 → 29, zero vacuous, zero false-green) · `cutover_ledger_gate` READY.

## Known gaps, stated not buried

- P7-PAR's human dashboard click-through was **not** performed (automated webui tests only)
- three oracle scenarios (`idle`, `drip`, `refresh`) were never captured; their Node behaviour is now unrecoverable
- P2-GOLD's live Node-vs-Kotlin differential is permanently impossible — the harness never existed on any ref, and the comparand is deleted
- the reasoning mirror is a no-op on the codex head, so "mirror visible" is unproven from its traffic
- `codex.rate_limits` never appeared in ~30k WS rounds, so the statusline soft-warn stays stale on a WS-only head


gateway/app/src/main/kotlin/splice/app/TurnPathProbeLoop.kt
```diff
@@ -0,0 +1,86 @@
...
+        conn.connectTimeout = timeoutMs
+        conn.readTimeout = timeoutMs
+        conn.doOutput = true
+        conn.setRequestProperty("Content-Type", "application/json")
+        conn.outputStream.use { it.write(PROBE_BODY) }
+        conn.responseCode // blocks up to readTimeout; a wedge never answers
+        conn.disconnect()
+        true
+    } catch (ignored: java.io.IOException) {
+        false
```
marcospaulo: **[concern]** The probe's contract ("ANY HTTP status is life; only a hang/timeout is death") is a good shape, and the bounded connect/read timeouts make the wedge case testable. At the posted head, though, `probeOnce()` catches only `java.io.IOException` (gateway/app/src/main/kotlin/splice/app/TurnPathProbeLoop.kt:76-78), so a runtime throwable out of HttpURLConnection (SecurityException, IllegalStateException) cancels the Dispatchers.IO job and nothing supervises or restarts it: the failure counter freezes and /health keeps reporting ok:true while the turn path is dead.

Your unpushed review-round fixes already address this (the supervise()/invokeOnCompletion wrapper that fails toward alarm). Could you push them? The PR head is what gets reviewed and merged, and right now it still carries the unsupervised version.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesStreamTranslator.kt
```diff
@@ -310,15 +386,32 @@ private class ResponsesEventReducer(val ctx: StreamTurnContext) {
     val blocks = HashMap<Int, BlockState>()
     var hasToolUse = false
     var emittedText = false
+
+    /** CX-09: a thinking block actually reached the sink. NOT the same as thinkingBuf being
+     *  non-empty — [harvestFallback] refills that buffer from the completed response object
+     *  without emitting anything, and that is precisely the turn the honesty gate must still
+     *  call empty. */
+    var emittedThinking = false
```
marcospaulo, line 394: **[concern]** The CX-09 split between `thinkingText` and `emittedThinking` is well documented here, and the harvest-fallback distinction it encodes is the right one. One gap on the redacted path: `emitReplayedReasoning` writes a client-visible block via `sink.addRedactedThinking(envelope)` (gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/ResponsesStreamTranslator.kt:698) but never sets `emittedThinking = true`, the way the native thinking paths do at lines 587 and 627. `TurnPipeline.nothingReachesTheClient` (gateway/gateway/src/main/kotlin/splice/gateway/pipeline/TurnPipeline.kt:131) reads only that flag and `willMirrorHere`, so a turn whose sole client-visible output is an encrypted reasoning envelope arrives with `emittedThinking=false` and `thinkingText` empty, gets classified as empty, and the client receives an `empty_model` api_error for a turn that actually put a block on the wire.

Setting the flag beside the sink write closes it:

```kotlin
if (shouldEmitReasoning(ctx, item)) {
    sink.addRedactedThinking(envelope)
    emittedThinking = true // CX-09: a redacted block reached the sink too
}
```

Reachability needs an encrypted-reasoning-only turn, which I have not seen in production traffic, so this reads as a correctness gap rather than an active incident. A reducer test driving only a redacted-reasoning item and asserting `Success.emittedThinking` would pin the contract.

gateway/app/src/main/kotlin/splice/app/TurnPathProbeLoop.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val log: (String) -> Unit,
+    private val intervalMs: Long = PROBE_INTERVAL_MS,
+    private val timeoutMs: Int = PROBE_TIMEOUT_MS,
+) {
+    private var consecutiveFailures = 0
+
+    public fun start(scope: CoroutineScope): Job = scope.launch(Dispatchers.IO) {
+        while (isActive) {
+            delay(intervalMs)
+            tick()
```
marcospaulo: **[concern]** The loop mirrors AuthProbeLoop's structure closely, which makes the one behavioral difference easy to trip over at the posted head: `delay(intervalMs)` runs before the first `tick()` (gateway/app/src/main/kotlin/splice/app/TurnPathProbeLoop.kt:38-41), so a head wedged on boot stays invisible for roughly one interval plus one timeout, where AuthProbeLoop probes immediately.

I understand your unpushed review-round commits keep this behavior and document it as deliberate (a boot-time false alarm judged worse than a 60s detection floor). Please push them. Without the rationale in the tree, this reads as an accidental divergence from the loop the file says it mirrors.

gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/Harvested.kt
```diff
@@ -120,18 +121,15 @@ private const val TYPE_TOOL_SEARCH_CALL = "tool_search_call"
...
+    val cached = details?.firstLong("cached_tokens")?.takeIf { it > 0 }
+        ?: usage.firstLong("cache_read_input_tokens") ?: 0L
     // output_tokens_details.reasoning_tokens carries the 518n-2 truncation fingerprint; absent on
     // non-reasoning backends (→ 0 → never fold).
-    val reasoning = (usage["output_tokens_details"] as? JsonObject)?.let { num(it, "reasoning_tokens") } ?: 0L
+    val reasoning = (usage["output_tokens_details"] as? JsonObject)?.firstLong("reasoning_tokens") ?: 0L
     return Usage(
-        inputTokens = num(usage, "input_tokens", "prompt_tokens"),
-        outputTokens = num(usage, "output_tokens", "completion_tokens"),
+        inputTokens = usage.firstLong("input_tokens", "prompt_tokens") ?: 0L,
```
marcospaulo, line 131: **[suggestion]** The dual-spelling read with `firstLong("input_tokens", "prompt_tokens")` (gateway/dialect-openai-responses/src/main/kotlin/splice/dialect/responses/Harvested.kt:131-132) is compact, and canonical-first precedence is the right order. The tests only exercise the canonical spellings, though: every usage fixture in ResponsesStreamTranslatorTest uses input_tokens/output_tokens, so a regression that dropped the alias fallback or flipped the precedence would stay green.

One fixture using prompt_tokens/completion_tokens, plus a case where both spellings are present and the canonical value wins, would pin the contract. This matters because these counts feed the context-window percentage downstream.

gateway/app/src/main/kotlin/splice/app/cli/AdminSupport.kt
```diff
@@ -122,7 +128,17 @@ internal object AdminSupport {
...
+                // JW-01: the spawned JVM's output lands in daemon-boot.log (rolled at 1MB, one
+                // generation), never /dev/null; an unwritable logs dir degrades the redirect
+                // instead of breaking the launch. Mirrors bin/splice-launch byte-for-byte in
+                // behaviour — the two cold-start paths must not drift.
+                "L='${StatePaths().logsDir}'; " +
+                    "B=\"\$L/daemon-boot.log\"; mkdir -p \"\$L\" 2>/dev/null; " +
+                    "[ -f \"\$B\" ] && [ \"\$(wc -c <\"\$B\" 2>/dev/null || echo 0)\" -gt 1048576 ] " +
+                    "&& mv -f \"\$B\" \"\$B.1\" 2>/dev/null; " +
+                    "if ( : >>\"\$B\" ) 2>/dev/null; then " +
+                    "nohup java \${SPLICE_JVM_OPTS:-$DEFAULT_JVM_OPTS} -jar '$jar' daemon >>\"\$B\" 2>&1 & " +
```
marcospaulo: **[concern]** Mirroring bin/splice-launch byte-for-byte is a defensible goal; the two cold-start paths drifting would be worse than what either does alone. One robustness gap in the string-built command: `$jar` is interpolated into a single-quoted literal inside the `sh -c` string (gateway/app/src/main/kotlin/splice/app/cli/AdminSupport.kt:140), so an install path containing an apostrophe (a home directory like `/home/o'brien` is enough) breaks out of the quoting and the cold start dies on a shell parse error.

To be precise about scope: the `SPLICE_JVM_OPTS` half does not have this problem, because the shell expands it after parsing and never re-tokenizes the value as commands. Only the path quoting is load-bearing here. Passing the jar as data instead of interpolating it removes the class:

```kotlin
ProcessBuilder(
    "sh", "-c",
    "nohup java \${SPLICE_JVM_OPTS:-$DEFAULT_JVM_OPTS} -jar \"\$1\" daemon >>\"\$B\" 2>&1 &",
    "sh", jar.toString(),
)
```

or build the java argument list directly and drop the shell for that invocation entirely.

gateway/app/src/main/kotlin/splice/app/cli/DoctorRuntime.kt
```diff
@@ -0,0 +1,84 @@
...
+ *  unreachable endpoint each degrade to one INFO row, never a crash and never a fabricated OK. */
+internal fun runtimeChecks(snapshot: DaemonSnapshot, envReader: (String) -> String?): List<DoctorCheck> {
+    val statePaths = StatePaths(envReader = envReader)
+    val skip = when {
+        !snapshot.running -> "skipped (daemon stopped)"
+        readMgmtKey(statePaths) == null -> "skipped (mgmt-key unreadable)"
+        else -> null
+    }
+    if (skip != null) return listOf(DoctorCheck("runtime", CheckStatus.INFO, skip))
+    val heads = ControlPlaneClient.headsRuntime(snapshot.port, checkNotNull(readMgmtKey(statePaths)))
```
marcospaulo: **[concern]** The fail-open shape of this section is clearly deliberate (every hop degrades to one INFO row, documented at gateway/app/src/main/kotlin/splice/app/cli/DoctorRuntime.kt:21), and the skip guard handles the unreadable-key case cleanly. The guard and the use read the key independently, though: line 26 checks `readMgmtKey(statePaths) == null`, then line 30 calls `readMgmtKey(statePaths)` again under `checkNotNull`. If the file empties out between the two reads (rotation, operator action), `checkNotNull` throws and `guarded` converts it to a FAIL row, contradicting the documented contract that an unreadable key degrades to INFO.

Caching the first read keeps the contract:

```kotlin
val key = readMgmtKey(statePaths)
val skip = when {
    !snapshot.running -> "skipped (daemon stopped)"
    key == null -> "skipped (mgmt-key unreadable)"
    else -> null
}
if (skip != null) return listOf(DoctorCheck("runtime", CheckStatus.INFO, skip))
val heads = ControlPlaneClient.headsRuntime(snapshot.port, checkNotNull(key))
```

webui/src/entities/config/api/index.ts
```diff
@@ -2,10 +2,10 @@ import { control } from '@shared/api';
 import type { ConfigValue, PatchResult } from '@shared/api';
 import { configStore } from '../model/store';
 
-export async function fetchConfig(): Promise<void> {
+export async function fetchConfig(head?: string): Promise<void> {
```
marcospaulo, line 5: **[concern]** The head-scoped fetch threads through cleanly here, and the page's `[head]`-keyed refetch keeps the selector and the store in sync on selection changes. One caller misses the new parameter: `applyConfigPatch` (webui/src/entities/config/api/index.ts:18) still calls `fetchConfig()` bare, so after a PATCH made while viewing a specific head, the store repopulates with the GLOBAL effective/layers view while the selector keeps showing that head. The page then renders global data under a per-head label until the user reselects.

Threading the current head through the patch refresh (or resetting the selector to 'global' on patch) keeps the two consistent:

```typescript
export async function applyConfigPatch(
  patch: Record<string, ConfigValue>,
  head?: string,
): Promise<PatchResult> {
  const result = await control.patchConfig(patch);
  await fetchConfig(head);
  return result;
}
```

dev/campaigns/proxy-hardening/oracle/replay.mjs
```diff
@@ -0,0 +1,425 @@
...
+function isSanctioned(entry, sanctioned) {
+  return sanctioned.some((s) => {
+    const leaf = s.field.split('.').pop();
+    if (!entry.path.endsWith(`.${leaf}`)) return false;
+    if (s.pinnedValue !== undefined) return JSON.stringify(entry.obs) === s.pinnedValue;
+    return s.without === undefined || entry.obs === s.without;
+  });
+}
+
+function post(port, body, bearer, path = '/v1/messages') {
```
marcospaulo, line 227: **[suggestion]** The replay harness reproduces the frozen fixtures faithfully, and the no-session-header choice is documented where it matters. One operational gap: `post()` (dev/campaigns/proxy-hardening/oracle/replay.mjs:182) issues the request with no timeout or AbortController, so if the Kotlin daemon wedges mid-scenario (exactly the failure class this oracle exists to catch) the replay hangs until the CI job kills it, and the failure surfaces as an opaque job timeout instead of a named harness error.

A per-request timeout would fail fast with a diagnosable message:

```javascript
req.setTimeout(30_000, () => req.destroy(new Error(`replay timeout on ${path}`)));
```

marcospaulo (commented):
Review of: Close the W4 hardening campaign on the Kotlin gateway (CX-02/CX-09/CX-11/CX-18 dialect fixes plus ~14 new campaign wall scripts) and delete the entire legacy Node stack (server/, 43 files, -6241 lines) now that the Kotlin daemon serves production alone, rewiring workspaces, gates, hooks, bin, and docs to the Kotlin-only world.

Findings: 2 high, 8 medium, 1 nit. (147 raw findings dissolved by evidence gates.)

Checked clean: api-abi, build-config, mechanical.

Optional follow-up: draft ast-grep rules for recurring classes (F141, F149) are available on request.

Additional observations:
- [2026-08-12] hif.py dedupe bug (file+line fallback ate cross-category findings) fixed mid-run; HIF rebuilt deterministically from on-disk inputs
- [2026-08-12] flowmap: WsRoundDriver seam continues out of candidate scope (WsSession.completed/cleared)
- [2026-08-12] flowmap: addRedactedThinking client-visibility VERIFIED at SseEmitter.kt:211-219 by orchestrator
- [2026-08-12] correctness-walls: CX-19 oracle-replay wall: the wall script itself is correctly fail-closed (selftest passes), but the oracle/fixtures/ directory was not materialized in this review working tree, so the wall reports NO FIXTURE for all 11 scenarios and ORACLE LOST. I cannot verify whether the 11 byte-exact fixtures exist in the real PR; the PR intent map lists this claim as claimed_but_unverifiable (needs a run).
- [2026-08-12] correctness-walls: All other 24 campaign wall scripts pass their --selftest synthetic controls and report GREEN against the PR-head Kotlin sources in this working tree, so their substring/call-site pins are aligned with the current implementation.
- [2026-08-12] correctness-walls: capture.mjs is inoperable-by-design after server/ deletion and exits 2 with a clear message; its early guard checks server/src/codex-proxy.mjs while the script conceptually drives server/test/codex-proxy.test.mjs, but in practice both files are deleted together.
- [2026-08-12] correctness-walls: webui P7-PAR click-through gap: the per-head selector is wired through fetchConfig(head), but applyConfigPatch drops the head, which is the one behavior a manual dashboard click-through would catch and an isolated entity test likely will not.
- [2026-08-12] api-abi: Several other newly-added public constructors and top-level functions were already flagged by the mechanical no-lambda-seam rule (e.g. ControlServer's topologyStale/turnPathStalled constructor lambdas, HeadScopedLog, CredentialLock.withLock, MgmtKey). They share the same raw-function-type ABI concern as AuthProbeLoop, but they either live in SPI/internal modules or duplicate the style rule, so I did not elevate them as separate API/ABI defects. If any are intended to be long-lived public API they should adopt named fun interfaces.
- [2026-08-12] api-abi: ResponsesRequest and ProviderConfig also grew new nullable serializable fields. Those changes are wire-compatible, but they still churn the generated copy()/componentN() contracts of public data classes. Worth confirming the team accepts the ABI cost before more knobs land on these DTOs.
- [2026-08-12] security: AdminSupport.openUrl passes topology/config-derived URLs straight to 'open' / 'xdg-open' without scheme validation; config-controlled so risk is low, but a malicious custom scheme or file:// URL could launch an unexpected handler.
- [2026-08-12] security: bin/splice-launch line 148 repeats the same unquoted SPLICE_JVM_OPTS expansion in the degraded-redirect branch; the same fix should cover both spawn paths.
- [2026-08-12] security: No .github/workflows, package.json dependency, or CI trigger changes appeared in the skim, so the workflow-trigger and dependency-pinning parts of the charter had no material to review.
- [2026-08-12] tests: CX-09 empty-turn honesty is pinned in PassthroughStreamTranslatorTest, not in ChatStreamTranslatorTest/ResponsesStreamTranslatorTest; the stream translator unit tests exercise reasoning alongside text but never a thinking-only turn, so they would not fail if the translator regressed on emittedThinking. The failure mode is covered, but not by the two files the prompt frames as primary evidence.
- [2026-08-12] tests: CX-11 is_error parsing lives in LoopGuard and AnthropicRequest parsing, tested by LoopGuardTest and AnthropicParseTest; the stream translator tests do not (and should not) cover it, so the prompt's framing of the two stream-translator tests as primary CX-11 evidence is off.
- [2026-08-12] tests: mech-tests flagged runBlocking and Thread.sleep in CredentialLockTest, WatchdogTest, HeadServerCapacityTest, HeadServerFoldTest, and MockChatGptUpstream. These are timing-sensitive integration tests using real file locks, real network, or real Netty warmups; the rule hits are style-level and I see no concrete bug beyond the rule text.
- [2026-08-12] build-config: Deleted-rule completeness check: exactly six .rules/rules/*.yml JS rules and six .rules/rule-tests/*-test.yml files are removed in the diff; no surviving build/config file references their ids or the .rules/rules/ directory.
- [2026-08-12] build-config: No active build/config file still references server/ as a runtime path; the remaining mentions in .rules/README.md, bin/claudex, and test_orchestrator.py are historical comments documenting the cutover.
- [2026-08-12] build-config: .rules/README.md lines 22-35 list six Kotlin/webui rules (kt-no-quality-suppress, kt-no-stream-options-request, kt-no-request-body-gzip, webui-fetch-only-in-api, webui-no-emdash-ui-text, webui-css-tokens-only) that do not exist on disk; only kt-head-log-prefix.yml is present. Those inventory lines are not added in this PR, but the PR's update to the README (lines 48-50) claims Kotlin successors live in .rules/kotlin-splice/ without updating the stale table.
- [2026-08-12] build-config: package.json and package-lock.json workspaces were cleaned to ["webui"]; no @splice/server or server/ package entries remain in the lockfile.
- [2026-08-12] security: AdminSupport.openUrl passes topology/config-derived URLs straight to 'open' / 'xdg-open' without scheme validation; config-controlled so risk is low, but a malicious custom scheme or file:// URL could launch an unexpected handler.
- [2026-08-12] security: bin/splice-launch line 148 repeats the same unquoted SPLICE_JVM_OPTS expansion in the degraded-redirect branch; the same fix should cover both spawn paths.
- [2026-08-12] security: No .github/workflows, package.json dependency, or CI trigger changes appeared in the skim, so the workflow-trigger and dependency-pinning parts of the charter had no material to review.
- [2026-08-12] correctness-dialects: CX-02 compaction directive symmetry: all three dialect builders now route compact-turn system text through CompactInstructions.withCompactDirective; the claimed 'Responses bytes sha-identical' is not independently verifiable from the diff alone.
- [2026-08-12] correctness-dialects: CX-11 is_error parsing: LoopGuard correctly prefers ToolResultBlock.isError and falls back to ERROR_MARKER only when isError is null, matching the intent to avoid overriding an explicit false.
- [2026-08-12] correctness-dialects: CX-18 usage alias chain: JsonScalars.firstLong is consumed by the intended readers (chat translator, passthrough translator, Responses harvest, UsageHud) and no additional hand-rolled fourth reader was found.
- [2026-08-12] correctness-dialects: CX-09 propagation: emittedThinking is preserved through FoldRunner/Search partial salvages and merged across rounds in TurnDriver; the missing source update is the only gap.
- [2026-08-12] correctness-dialects: TurnPipeline imports splice.core.turn.pickModelText and ResponsesStreamTranslator imports splice.core.turn.isWeakSummaryText, but neither function appears in the provided changed core/turn files; assuming they are unchanged pre-existing code.
- [2026-08-12] RELIEF_AUDIT (orchestrator): scanned all 147 dissolutions; the 7 lens dissolutions carry concrete checkable evidence (no published artifacts, shell expansion semantics, orchestrator-verified file existence); no relief fingerprints; subagent's re-run of F146/F147/F148 stood. Nothing flips.

gateway/app/src/main/kotlin/splice/app/cli/ControlPlaneClient.kt
```diff
@@ -30,16 +79,59 @@ internal object ControlPlaneClient {
...
+        // daemon nobody had asked to stop, then told the operator to "terminate manually"
+        // (observed twice on 2026-08-11).
         runCatchingCancellable {
-            request("http://127.0.0.1:$port/api/daemon/shutdown", method = "POST", bearer = key) { true }
-        }.discard("POST may fail on graceful teardown; the stop poll below is the real stop signal")
-        repeat(STOP_POLLS) {
+            request("http://127.0.0.1:$port/api/daemon/shutdown", method = "POST", bearer = key) { conn ->
+                if (conn.responseCode == HTTP_UNAUTHORIZED || conn.responseCode == HTTP_FORBIDDEN) {
+                    println(
+                        "splice: shutdown request REJECTED (${conn.responseCode}) — the mgmt key on disk " +
```
marcospaulo: **[concern]** Surfacing an auth rejection on the shutdown POST instead of discarding it is the right call; the comment's 2026-08-11 incident note shows exactly why the silent case is expensive. At the posted head the message is unreachable, though: `request()` only invokes the read lambda for 2xx responses (gateway/app/src/main/kotlin/splice/app/cli/ControlPlaneClient.kt:164-165), so on a 401/403 it returns null before the rejection branch at :89-94 runs, and the operator never sees the reason before the signal ladder engages.

Your unpushed review-round fixes already make the rejection visible (statusOf no longer gates on 2xx). Could you push them? The posted head still has the dead branch.

gateway/app/src/main/kotlin/splice/app/cli/ControlPlaneClient.kt
```diff
@@ -30,16 +79,59 @@ internal object ControlPlaneClient {
...
         return stopped(port)
     }
 
+    /** Every live splice DAEMON process, identified by cmdline. Matches both the installed jar
+     *  (splice.jar) and a dev-tree jar (app-all.jar); never anything else — the same scoping that
+     *  keeps the oracle harness's self-heal away from processes it does not own. */
+    private fun spliceDaemons(): List<ProcessHandle> = ProcessHandle.allProcesses()
+        .filter { ph ->
+            val cmd = ph.info().commandLine().orElse("")
+            cmd.contains("java") && cmd.contains("daemon") &&
```
marcospaulo: **[concern]** Scoping the stop ladder to splice daemons is the right instinct, and the comment about keeping the oracle harness's self-heal away from unowned processes shows the threat shape was considered. At the posted head the ownership proof is three cmdline substrings (`java`, `daemon`, `splice.jar`/`app-all.jar`, gateway/app/src/main/kotlin/splice/app/cli/ControlPlaneClient.kt:127-133), so any process whose command line happens to contain those tokens (a renamed dev wrapper, an unrelated Java service) gets SIGTERM and then SIGKILL.

Your unpushed review-round fixes already scope kills to the process holding the target port. Could you push them? The posted head still carries the substring match.

gateway/gateway/src/test/kotlin/head/HeadServerCapacityTest.kt
```diff
@@ -197,4 +198,22 @@ class HeadServerCapacityTest {
         mock.releaseHold()
         held.await()
     }
+
+    @Test
+    fun `restart clears an armed 429 cooldown`() = runBlocking {
+        // NF-01: an upstream 429 arms the head-wide fail-fast horizon on the long-lived
+        // UpstreamClient; a control-plane restart must be a real escape hatch, not a no-op.
+        client.post("http://127.0.0.1:$port/v1/messages") {
```
marcospaulo, line 206: **[nit]** The restart-clears-cooldown scenario covers the NF-01 escape hatch well. Small diagnosability point (preference, not requirement): the test asserts `rateLimitedForMs > 0` without first asserting the post returned 429 (gateway/gateway/src/test/kotlin/head/HeadServerCapacityTest.kt:206), so if the quota429 mock scenario ever drifts, the failure reads identically to a broken cooldown. Capturing the response and asserting status 429 as a precondition would make the two failures distinguishable.

PR #109, opened 2026-09-01
fix(gateway): stall watchdog keys on the first client frame, compact turns get the total cap, policy refusals are terminal

marcospaulo:
## Why

Codex compaction has been failing across projects since this morning. `~/.claude-codex/claudex-compact-stats.jsonl` for 2026-09-01: 30 `model_text` vs **153 `stream_error`** (109 `overloaded_error`, 44 `api_error`), against 1–2 failures/day for the previous week. Two distinct mechanisms, both ours to fix at the proxy:

### 1. The stall watchdog reaped silent compactions on the wrong tier

Every one of the 109 stalls reads `no completion within the 180s idle cap` with `first_byte=1–5s`, `events_in=5`, **no first delta**, on 1.0–1.2MB upstream bodies. The successful compactions the same morning had first deltas at 17s, 19s and 111s — the same silence, one tier apart.

`TurnWatchdog` had two tiers by design (the v35 spec in `Watchdog.kt`): `firstByteTimeoutMs` (300s) until the first byte, `streamIdleMs` (180s) after it, precisely so a big-context prefill "legitimately silent for minutes" is not reaped. But the tier flipped on the first upstream **byte** (`markByte()` on the raw SSE read / the first WS event), and on the Responses API that byte is the `response.created` handshake, which arrives in 1–5s while the model is still reading the prefill. So every large compaction was pinned to the short tier before a single token existed, aborted at ~180s, surfaced as a retryable `overloaded_error`, re-sent cold by Claude Code, and looped (three sessions in parallel, `inflight=3`). The `drift-repair` ledger's DR-7 notes record the same fact ("TearAwareEvents.onBytes calls markByte on the RAW read, so response.created flips the watchdog to streamIdle before the stall begins") but the `SseRoundConsumeTest` arm pinned that behaviour as correct.

**Fix:** the tier now reads the round's own `ClientFrameEmitted` probe (CONTENT_FRAMES_OUT above the round baseline — the same fact G5 keys reissue on), on both transports. Until the client has been handed content, silence is judged on `firstByteTimeoutMs`; after it, on `streamIdleMs`. `markByte()`/`resetFirstByte()` are gone (`resetRound()` keeps only the DR-7 stale-Idle clear); bytes still touch the slot for liveness. The `WatchdogFired.Idle` sentinel now carries `sawClientFrame` and `limitMs`, and the terminal message names the tier that fired instead of always saying "180s idle cap".

### 2. Content-policy refusals were retried as if transient

44 compact turns (and 200 normal turns) today failed with `ChatGPT backend: cyber_policy This content was flagged for possible cybersecurity risk…` — OpenAI's Trusted-Access classifier on the whole context (a known false-positive class: openai/codex#37473, #34864, #41672). DR-72 stopped mislabelling it as `authentication_error`, but its `api_error` verdict is still retryable to Claude Code, so each refusal became a ~30s × N backoff storm of the same 1.3MB transcript, and a compaction that trips it can never complete. The vendor itself returns this refusal as HTTP 400 / `invalid_request` when it arrives pre-stream.

**Fix:** the documented prompt-refusal codes (`cyber_policy`, plus the Responses API enum's `invalid_prompt`, `bio_policy`, `image_content_policy_violation`) classify as `INVALID_REQUEST` — terminal to the client, vendor remedy text intact, provenance by exact code (same rule as `RETRYABLE_CODES`). Nothing splice can do makes the classifier accept the content; what it can do is stop paying for the same rejection ten times and let the user switch heads for the compaction.

### 3. A compact turn's pre-output silence is bounded by the whole-turn cap only (second commit)

Deployed at 13:58 with fixes 1 and 2, the next three compactions (14:07:59, 14:13:50, 14:14:41) died on the now-correct tier: `no first output within the 300s first-output cap` — first byte at 4–6s, `events_in=5`, 1.1–1.2MB upstream bodies, ~350s of reasoning silence before the model would have started the summary. A compaction is one long reasoning pass over the whole transcript; the only pre-output wall that makes sense for it is the whole-turn wall.

**Fix:** `WatchdogBudget.forCompact()` copies the budget with `firstByteTimeout = totalCap`; `TurnDriveFactory` selects it when the turn is a compaction. Normal turns keep the 300s tier, and `streamIdleMs` still applies to every turn once output flows.

## Evidence-first

Red on unmodified `feat/claude-head` (26a9b9b2), then green:

- `UpstreamFailurePolicyRefusalTest`: `a cyber_policy refusal is a terminal invalid_request…` and `every documented prompt-refusal code…` — RED (`expected: <INVALID_REQUEST> but was: <API_ERROR>`).
- `SseRoundConsumeTest` `a pre-content idle reap is an outcome, not a transport tear - DR-7`: budgets swapped to 1s first-output / 20s streamIdle — RED (`expected an Idle reap: null`: the ack flipped the tier and the 20s streamIdle let the mock hang up first).
- `WatchdogTest`: new arm `a handshake byte before any client frame keeps the first-output tier` pins the mechanism at the unit level (touch without a frame → not reaped past 3× streamIdle; then the first frame → reaped on streamIdle with `limitMs` named).
- `WatchdogBudgetTest` (core): `forCompact()` moves only the first tier. `HeadServerCompactCapTest` (gateway, real `HeadServer` + `CodexProvider` against the mock upstream's `idlepre` scenario): a normal turn silent after the handshake is reaped by a 1s first-output cap; the same silence on a compact turn survives it and is ended only by the 4s whole-turn wall.

## Test robustness (commits 3 and 4, test-only)

- `HeadServerCompactCapTest` builds one head per arm: the normal arm's turn spans two upstream attempts (WS, then the SSE re-serve), so it no longer shares the compact arm's 4s whole-turn wall (the kover-instrumented coverage job on 2d9e2c0a lost that race; the gate job passed).
- `HeadServerIntegrationTest`'s two log-reading tests wait for their own turn's perf line under the list's lock: `TurnFinish` sends the terminal frames first (DR-129) and records perf last, so reading `logs` the moment the client body completes raced the server thread (coverage job on 1a2e81c7; the gate job passed, and the previous fifteen coverage runs on `feat/claude-head` were green).

## Not changed

- No knob defaults moved. `firstByteTimeoutMs` (300s) is the cap a silent NORMAL turn gets (Codex CLI's own `stream_idle_timeout_ms` default is 300000); a silent compaction gets `upstreamTimeoutMs` (900s on claudex). Successful compactions have run to 733s; if gpt-5.6 xhigh needs longer on a 250k transcript, `[heads.claudex.overrides] upstreamTimeoutMs` is the per-head override (kimi already carries 2400000), not a code change.
- The WS→SSE re-serve on a pre-frame failure terminal still runs for a policy refusal (two upstream attempts instead of ~20). Skipping the re-serve for a deterministic code is a follow-up.
- Ledger: this branch's `drift-repair.toml` is owned by the claude-splice seat per its 2026-08-30 two-seat law; no row was written here. Suggested row: the DR-7 arm C2 note above, closed by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,20 @@
 
 ### Fixed
 
+- **Codex compaction no longer dies at the idle cap while the model is still reading.** The stall
+  watchdog switched to its short `streamIdleMs` tier on the first upstream byte, and on the
+  Responses API that byte is the `response.created` handshake, not output — so a compaction that
+  reasoned silently over a large transcript for longer than 180s was aborted and re-sent cold by the
+  client, in a loop (109 stalls on one head in a single day). The tier now follows the first client
+  content frame: until the client has seen output the silence is judged on `firstByteTimeoutMs`,
+  after it on `streamIdleMs`, on both the SSE and WebSocket transports. The stall message names the
+  tier that actually fired.
+- **A content-policy refusal is terminal, not retried.** ChatGPT's `cyber_policy` flag (and the
+  Responses API's documented prompt refusals: `invalid_prompt`, `bio_policy`,
+  `image_content_policy_violation`) reached Claude Code as a retryable `api_error`, so every refusal
+  became a backoff storm of the same multi-megabyte transcript. They now surface as
+  `invalid_request_error` with the vendor's own remedy text, matching the HTTP 400 the vendor returns
+  for the same refusal pre-stream.
 - Refresh failures for Codex, Grok, and Kimi no longer risk logging vendor response bodies, and
   KeyStore values containing `#`, quotes, or backslashes round-trip without corruption.
 - Request-body torn wakeups become an Anthropic-shaped HTTP 400 without swallowing genuine coroutine
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,9 @@
   reasoned silently over a large transcript for longer than 180s was aborted and re-sent cold by the
   client, in a loop (109 stalls on one head in a single day). The tier now follows the first client
   content frame: until the client has seen output the silence is judged on `firstByteTimeoutMs`,
-  after it on `streamIdleMs`, on both the SSE and WebSocket transports. The stall message names the
-  tier that actually fired.
+  after it on `streamIdleMs`, on both the SSE and WebSocket transports. A compact turn's pre-output
+  silence is bounded by `upstreamTimeoutMs` alone (compactions on the corrected tier still
+  died silent at the 300s cap). The stall message names the tier that actually fired.
 - **A content-policy refusal is terminal, not retried.** ChatGPT's `cyber_policy` flag (and the
   Responses API's documented prompt refusals: `invalid_prompt`, `bio_policy`,
   `image_content_policy_violation`) reached Claude Code as a retryable `api_error`, so every refusal
--- a/gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
+++ b/gateway/core/src/main/kotlin/splice/core/turn/Turn.kt
@@ -200,11 +200,21 @@ public class SharedSummaryParts(
     }
 }
 
-/** The two-tier watchdog knobs (v35 doctrine): before first byte the idle limit is
- *  firstByteTimeout (prefill is legitimately silent for minutes); after, streamIdle;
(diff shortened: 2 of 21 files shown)


marcospaulo:
Merging on the green gate (all ten checks pass on 553b09e4; local full gate green). Both review seats (codex-splice, claude-splice) are no longer running and the operator has ordered the release today, so this merge records no peer review.

PR #115, opened 2026-09-02
fix(claudex): retry the backend capacity signal and let Claude Code outlive the proxy wall

marcospaulo:
## What broke (read off the live daemon, 2026-09-01)

Every claudex compaction today was a 1.2 MB request that took 500 to 580 s when it worked. Two of them died as `client_abort` at 600 s and 604 s, one died in 7.5 s on the backend's `server_is_overloaded`, and from 21:01 one session retried its compaction every ~300 s with the daemon recording nothing.

## Fix 1: overload is classified by code shape, and it is `overloaded_error`

The in-stream `response.failed` carried code `server_is_overloaded`. The classifier trusts only an exact allowlist of codes, so this became a non-transient `api_error`: no reissue, no salvage, and Claude Code reported the compaction failed. codex-rs (PR #31058) names `server_is_overloaded` and `slow_down` as the backend's two capacity codes and retries them patiently. Any code spelling overload, plus `slow_down`, is now a transient `OVERLOADED` (the 529 / `overloaded_error` Claude Code retries with backoff). DR-71 had patched the same gap for `engine_/model_overloaded` by adding spellings; this replaces the list with one rule.

## Fix 2: the wrapper plants `API_TIMEOUT_MS` from the head's own wall

Claude Code's default request timeout is 600 s and it also bounds the first-byte deadline; the daemon's whole-turn cap is 900 s. Every compaction over ten minutes was killed by the client while the daemon was still streaming it. Each wrapper now launches Claude Code with `API_TIMEOUT_MS = upstreamTimeoutMs + 60 s` (960 s on defaults), derived in `LaunchSpecFactory` from the same `WatchdogBudget` the daemon enforces.

## Proof

- Red: the four new classifier arms fail on the old classifier (`expected OVERLOADED but was API_ERROR`), the launch-env arm fails without the planted variable (`expected 960000 but was null`).
- Green: 582 targeted tests, then the full `npm run gate` on this tip: `GATE: PASS` (first run failed only on `webui dist` from a stale local node_modules; `npm ci` and a rebuild reproduce the committed bundle byte for byte).

## Not fixed here, deliberately named

The ~300 s retry loop after 21:01 is a third defect: while an SSE upstream POST waits for response headers, the daemon writes nothing to Claude Code, so the client's first-byte deadline fires and the daemon records no verdict for the aborted turn. The evidence and the proposed change (commit headers and keepalive before the upstream handoff, and record client aborts before commit) are in the session notes; it needs its own red test against the mock upstream's stall scenario before it is touched.

## Update 2026-09-01 22:10 — the third symptom explained, build installed locally, Docker e2e green

**The ~300 s retry loop is Claude Code's non-streaming fallback, and both fixes above cover it.** Read from the installed binary (2.1.257): after a streaming attempt errors, Claude Code logs `Error streaming, falling back to non-streaming mode` and re-issues the query with `stream: false` and `timeout: API_TIMEOUT_MS || 300000`. The daemon serves `stream:false` on its collect path, which by design writes nothing until the terminal JSON body (no SSE channel, no keepalive), and a compaction's first upstream content arrives 500–580 s after `response.created` (perf rows today: `first_delta` 500502 / 576329 / 578572 ms). So every fallback attempt died client-side at 300 s with zero bytes on the wire, the daemon cancelled the upstream on the socket close (no perf row), and the client retried. Fix 1 removes the trigger in the observed case (`server_is_overloaded` no longer surfaces as a terminal `api_error`); fix 2 raises the fallback's timeout to 960 s, above the daemon's 900 s wall, so even a fallback compaction completes. `CLAUDE_CODE_DISABLE_NONSTREAMING_FALLBACK` exists in the client and would keep compaction on the streaming path (where the daemon does keep the client alive); not planted here — separate decision.

**Installed locally.** Jar `f80a075d59369ed9` (this branch at 3912a166, built through buildgate) installed via `install.sh` at 22:08 with the beta.1 jar backed up beside it; daemon restarted at inflight 0/0/0/0, 4/4 heads ready. The live `/launch/claudex` recipe now carries `API_TIMEOUT_MS=960000`.

**Fresh-machine Docker e2e** (harness in its own PR): this build 19/19; the v0.3.0-beta.1 assets 17/19, failing exactly the two launch-recipe steps (`API_TIMEOUT_MS` absent).

### DIFF
--- a/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
+++ b/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
@@ -137,6 +137,9 @@ public class LaunchService(
             // deadline; the head's whole-turn cap is the wall that decides a turn, so the client
             // must outlive it (spec.apiTimeoutMs = cap + grace). Without this every compaction
             // longer than ten minutes died as client_abort while the daemon was still serving it.
+            // Planted unconditionally: the head's own wall owns this deadline, so an ambient
+            // API_TIMEOUT_MS is replaced rather than merged (unlike NO_PROXY below); a larger value
+            // buys nothing past totalCap and a smaller one recreates the abort.
             put("API_TIMEOUT_MS", spec.apiTimeoutMs.toString())
             put("NO_PROXY", mergedNoProxy())
             // Hide Claude Code's built-in Anthropic-account commands: in a gateway head, auth is the
--- a/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
+++ b/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
@@ -140,6 +140,11 @@ public class LaunchService(
             // Planted unconditionally: the head's own wall owns this deadline, so an ambient
             // API_TIMEOUT_MS is replaced rather than merged (unlike NO_PROXY below); a larger value
             // buys nothing past totalCap and a smaller one recreates the abort.
+            // Claude Code's non-streaming fallback (after a streaming error it re-sends the turn
+            // with stream:false under this same deadline) stays ENABLED on purpose: the collect
+            // path answers it silently until the terminal body, and with the deadline past totalCap
+            // the head's wall speaks first. CLAUDE_CODE_DISABLE_NONSTREAMING_FALLBACK would only
+            // delete a recovery path.
             put("API_TIMEOUT_MS", spec.apiTimeoutMs.toString())
             put("NO_PROXY", mergedNoProxy())
             // Hide Claude Code's built-in Anthropic-account commands: in a gateway head, auth is the
--- a/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
+++ b/gateway/control/src/main/kotlin/splice/control/LaunchService.kt
@@ -133,6 +133,11 @@ public class LaunchService(
             put("CLAUDE_CODE_AUTO_COMPACT_WINDOW", max(AUTO_COMPACT_FLOOR, spec.contextWindow).toString())
             put("CLAUDE_AUTOCOMPACT_PCT_OVERRIDE", "85")
             put("MAX_THINKING_TOKENS", "128000")
+            // Claude Code's default request timeout is 600s and it also bounds the first-byte
+            // deadline; the head's whole-turn cap is the wall that decides a turn, so the client
+            // must outlive it (spec.apiTimeoutMs = cap + grace). Without this every compaction
+            // longer than ten minutes died as client_abort while the daemon was still serving it.
+            put("API_TIMEOUT_MS", spec.apiTimeoutMs.toString())
             put("NO_PROXY", mergedNoProxy())
             // Hide Claude Code's built-in Anthropic-account commands: in a gateway head, auth is the
             // proxy bearer above, so /login (a local-jsx command hardwired to platform.claude.com —
--- a/gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
+++ b/gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
@@ -232,11 +232,21 @@ public object UpstreamFailureClassifier {
 
     private sealed class ExtractResult {
         data class Fields(val message: String, val code: String) : ExtractResult() {
-            /** Overload: a 502 from the gateway, or capacity by CODE SHAPE (classifyContent's branch
-             *  comment carries the provenance). Lives on the fields, not the object: the object sits
-             *  at detekt's function budget and classifyContent at its complexity budget. */
-            fun isOverload(status: Int?): Boolean =
-                status == BAD_GATEWAY || code.lowercase() in CAPACITY_CODES || overloadCodeRe.containsMatchIn(code)
+            /** Overload: a 502 from the gateway, or capacity by CODE SHAPE on a status-less (in-stream)
+             *  or server-side failure (classifyContent's branch comment carries the provenance). A 4xx
+             *  keeps its deterministic verdict whatever its code spells: this arm sits ahead of
+             *  statusFallback's client-error branch, so without the floor a permanent 403 spelling
+             *  "overload" would be re-POSTed UPSTREAM_RETRIES times (review of #115). Lives on the
+             *  fields, not the object: the object sits at detekt's function budget and
+             *  classifyContent at its complexity budget. */
+            fun isOverload(status: Int?): Boolean = when {
+                status == BAD_GATEWAY -> true
+                status == null -> capacityShape()
+                else -> status >= SERVER_ERROR_FLOOR && capacityShape()
+            }
+
+            private fun capacityShape(): Boolean =
+                code.lowercase() in CAPACITY_CODES || overloadCodeRe.containsMatchIn(code)
         }
         data class Gateway(val failure: ClassifiedFailure) : ExtractResult()
     }
--- a/gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
+++ b/gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
@@ -148,7 +148,16 @@ public object UpstreamFailureClassifier {
                 ClassifiedFailure(ErrorType.RATE_LIMIT, msg.take(MAX_MESSAGE))
             status == AUTH_STATUS || authRe.containsMatchIn(blob) ->
                 ClassifiedFailure(ErrorType.AUTHENTICATION, msg.take(MAX_MESSAGE))
-            status == BAD_GATEWAY ->
+            // Overload: a 502 from the gateway, or capacity by CODE SHAPE. The ChatGPT backend
+            // reports "model at capacity" as HTTP 503 or an in-stream response.failed whose code is
+            // server_is_overloaded / slow_down — codex-rs names exactly those two (PR #31058, which
+            // turned them from turn-ending into a patient same-turn retry). An exact allowlist keeps
+            // losing this race one spelling at a time: DR-71 added engine_/model_overloaded, and on
+            // 2026-09-01 20:56 a claudex compaction (830KB upstream body) died as a non-transient
+            // api_error on server_is_overloaded — no reissue, no salvage, "compaction failed" in the
+            // client. Any code spelling overload IS the named transient server condition, and it
+            // surfaces as OVERLOADED (529 / overloaded_error): the type Claude Code retries on.
+            fields.isOverload(status) ->
                 ClassifiedFailure(ErrorType.OVERLOADED, msg.take(MAX_MESSAGE), transient = true)
             else -> statusFallback(status, msg, fields.code)
         }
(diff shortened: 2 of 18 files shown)


gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
```diff
@@ -222,7 +231,13 @@ public object UpstreamFailureClassifier {
     public fun mapOutStatus(status: Int): Int = if (status == BAD_GATEWAY) OVERLOADED_STATUS else status
 
     private sealed class ExtractResult {
-        data class Fields(val message: String, val code: String) : ExtractResult()
+        data class Fields(val message: String, val code: String) : ExtractResult() {
+            /** Overload: a 502 from the gateway, or capacity by CODE SHAPE (classifyContent's branch
+             *  comment carries the provenance). Lives on the fields, not the object: the object sits
+             *  at detekt's function budget and classifyContent at its complexity budget. */
+            fun isOverload(status: Int?): Boolean =
+                status == BAD_GATEWAY || code.lowercase() in CAPACITY_CODES || overloadCodeRe.containsMatchIn(code)
```
marcospaulo: **[concern]** The shape rule is the right instinct for the status-less case, and the cascade still puts overflow, policy refusal, 429 and 401 ahead of it, so the arms that have to win still win. One interaction moves with it that the description does not name. The old arm was `status == BAD_GATEWAY ->`; the new one is `fields.isOverload(status) ->` (UpstreamFailureClassifier.kt:160), and the predicate consults status only for the 502 equality:

```kotlin
fun isOverload(status: Int?): Boolean =
    status == BAD_GATEWAY || code.lowercase() in CAPACITY_CODES || overloadCodeRe.containsMatchIn(code)
```

Because that arm sits ahead of `else -> statusFallback(status, msg, fields.code)` (:162), a 4xx that is neither 401 nor 429 and carries a structured code containing "overload" (a 403 spelling it `overload_protection_denied`, say) no longer reaches statusFallback's CLIENT_ERROR_FLOOR branch (:221-222) as a terminal INVALID_REQUEST. It becomes OVERLOADED with transient=true, so the caller re-POSTs the whole context up to UPSTREAM_RETRIES (4) times with backoff against an account that may already be rate limited. Bounded rather than a storm, and the PR does name the retry cost of a permanent failure spelled with that word. What it does not name is that this arm also overrides the 4xx deterministic branch, not only the status-less RETRYABLE_CODES path the comments discuss.

If the status-less widening is the goal (it reads that way) and the 4xx override is incidental, gating the shape check on a server-side status leaves every currently pinned case green (503 plus a capacity code passes, every status-less SSE case passes) and changes only the untested branch:

```kotlin
fun isOverload(status: Int?): Boolean {
    val shape = code.lowercase() in CAPACITY_CODES || overloadCodeRe.containsMatchIn(code)
    return when {
        status == BAD_GATEWAY -> true
        status == null -> shape
        else -> status >= SERVER_ERROR_FLOOR && shape
    }
}
```

That holds at two conditions per branch, so it stays inside the budgets the current shape was written around.

If the override is intended (the backend does sometimes hand back capacity below 500), a test driving `status = 400, code = "server_is_overloaded"` pins that intent so the next reader does not "fix" it away. I may be wrong about how this backend spells capacity on a 4xx, since I have not seen one, and that is the argument for pinning the branch rather than reasoning about it.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
```diff
@@ -244,23 +259,25 @@ public object UpstreamFailureClassifier {
...
-        // DR-71: the engine/model-scoped overload spellings vendors emit are the same named
-        // transient server condition — a deterministic-looking absence here made them non-retryable.
-        "engine_overloaded",
-        "model_overloaded",
     )
 
+    // The backend's two capacity codes by name (codex-rs PR #31058); `slow_down` does not spell
+    // overload, so the shape rule alone would miss it.
+    private val CAPACITY_CODES = setOf("server_is_overloaded", "slow_down")
+    private val overloadCodeRe = Regex("overload", RegexOption.IGNORE_CASE)
```
marcospaulo, line 289: **[suggestion]** The new tests land where it matters: both capacity codes with a case variant, a structured code outranking deterministic-reading free text, and the pre-stream 503 shape. The one thing they do not pin is the widening itself. Delete the regex, hard-code the four spellings the tests actually use, and the whole suite stays green:

```kotlin
// mutation: every assertion in UpstreamFailureRetryWordingTest.kt still passes
private val CAPACITY_CODES = setOf(
    "server_is_overloaded", "slow_down", "engine_overloaded", "model_overloaded",
)
// private val overloadCodeRe = Regex("overload", RegexOption.IGNORE_CASE)
```

isOverload lowercases before the set check, so even the `SERVER_IS_OVERLOADED` case survives that mutation. A regression from the shape rule back into the exact-allowlist idiom would therefore ship green, and the allowlist is the thing that kept losing the race one spelling at a time. Two cases close it, and neither changes behavior:

```kotlin
@Test
fun `an unseen overload spelling is capacity by shape, not by allowlist`() {
    val r = coded("gpu_pool_overload_backpressure")
    assertEquals(ErrorType.OVERLOADED, r.type)
    assertTrue(r.transient)
}

@Test
fun `a 4xx carrying an overload-shaped code keeps the intended verdict`() {
    val r = UpstreamFailureClassifier.classify(
        FailureSource.HTTP,
        """{"error":{"code":"server_is_overloaded","message":"at capacity"}}""",
        status = 403,
    )
    // assert whichever way the precedence question is meant to go
}
```

The first is the regression guard for the rule this fix rests on (UpstreamFailureClassifier.kt:279). The second is the case the review priorities in the description ask about, and right now it has no answer in either direction.

gateway/control/src/main/kotlin/splice/control/LaunchService.kt
```diff
@@ -133,6 +133,11 @@ public class LaunchService(
             put("CLAUDE_CODE_AUTO_COMPACT_WINDOW", max(AUTO_COMPACT_FLOOR, spec.contextWindow).toString())
             put("CLAUDE_AUTOCOMPACT_PCT_OVERRIDE", "85")
             put("MAX_THINKING_TOKENS", "128000")
+            // Claude Code's default request timeout is 600s and it also bounds the first-byte
+            // deadline; the head's whole-turn cap is the wall that decides a turn, so the client
+            // must outlive it (spec.apiTimeoutMs = cap + grace). Without this every compaction
+            // longer than ten minutes died as client_abort while the daemon was still serving it.
+            put("API_TIMEOUT_MS", spec.apiTimeoutMs.toString())
```
marcospaulo, line 148: **[nit]** This answers the open question in the description ("does an operator-set API_TIMEOUT_MS get clobbered or respected?") in the clobbered direction, and clobbering looks correct to me, so this is only about saying so at the call site. Traced: the put at LaunchService.kt:140 is unconditional, envReader is consulted exactly once in the file (mergedNoProxy, :158-165, for NO_PROXY), API_TIMEOUT_MS is not in staleEnvUnsets (:90-111), and the shim execs `env` without -i (this file's header, :81-89), so an exported value really is present in the inherited environment and really is replaced on every /launch.

That reads as the contract rather than an oversight. LaunchTypes.kt:25-30 and LaunchSpecFactory.kt:66-68 derive the value from the same totalCap the watchdog enforces and refuse a second hand-maintained number, and the watchdog ends the turn at totalCap whatever the client's clock says. A larger operator value buys nothing, and a smaller one only recreates the client_abort this PR fixes. NO_PROXY is not a counter-precedent: a comma list composes with the operator's entries, a single scalar deadline has one owner.

What is missing is only that a reader has to run that trace to know it. Half a line on the comment already there:

```kotlin
// Planted unconditionally: the head's own wall owns this deadline, so an ambient
// API_TIMEOUT_MS is replaced rather than merged (unlike NO_PROXY below).
put("API_TIMEOUT_MS", spec.apiTimeoutMs.toString())
```

Preference, not a requirement. One thing I cannot close from the diff: bin/splice-launch is unmodified and out of scope here, so the "env without -i" step rests on this file's header comment rather than on the shim itself. If that ever became `env -i`, the ambient value stops existing and this note stops mattering.

marcospaulo (commented):
Review of #115: reclassify the ChatGPT backend's capacity failures (server_is_overloaded, slow_down, HTTP 502) as transient overloaded_error, and plant API_TIMEOUT_MS (daemon totalCap plus 60 s) into every wrapper's launch env so Claude Code's request cap outlives the daemon's 900 s wall.

Findings: 2 concerns (both on the widened overload arm in UpstreamFailureClassifier.kt: a 4xx carrying an overload-shaped code now bypasses the deterministic 4xx branch, and no test pins the shape rule against a regression to an exact allowlist), 1 nit (the unconditional API_TIMEOUT_MS write is the intended contract; say so at the call site). 4 raw findings dissolved by the evidence gates (a mechanical rule hit judged a false positive for this private nested data class, a changelog wording nit, and two cross-lens duplicates consolidated).

Checked clean: cascade ordering for overflow, policy refusal, 429 and 401 is unchanged; removing the overload spellings from RETRYABLE_CODES leaves no silent non-transient fallback (status 500/502/503 with those codes still reach the overload arm first); apiTimeoutMs derives from the same boot-time totalCap the live watchdog enforces (one construction site, restart-required knob, no split-brain); all four LaunchSpec constructor sites updated; new field has KDoc and follows the Ms naming; tests pin the exact env string ("960000") and both operands of the cap-plus-grace formula; no credential, injection or workflow surface touched.

Mode: compact (122 added lines, 9 files), one reviewer over every lens charter with the flow map folded in. Mechanical scan: run (94 rules from the bundled and repo packs, 1 hit, dissolved). Mechanical-only coverage: CHANGELOG.md, ControlServerTest.kt. No reduced-confidence findings. No antibody drafts offered.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
marcospaulo: Fixed in 6aab7134: isOverload now answers true for a 502, by shape when the status is null (in-stream), and by shape only at or above the 5xx floor otherwise. A 403 spelling overload keeps statusFallback's INVALID_REQUEST verdict, pinned by a new test.

gateway/provider-spi/src/main/kotlin/splice/spi/UpstreamFailureClassifier.kt
marcospaulo: Fixed in 6aab7134: added `an unseen overload spelling is capacity by shape, not by allowlist` (the allowlist mutation now ships red) and `a 4xx carrying an overload-shaped code stays a deterministic client error`. UpstreamFailureRetryWordingTest 13/13 green, detekt and the ast-grep walls clean.

gateway/control/src/main/kotlin/splice/control/LaunchService.kt
marcospaulo: Fixed in 6aab7134: the call-site comment now states that the unconditional write is the contract (the head's wall owns the deadline, so an ambient value is replaced rather than merged, unlike NO_PROXY).

marcospaulo:
Readiness-sweep follow-up, four more commits on this branch (each closes a residual risk from the sweep, all verified locally: `:core:test`, `:provider-spi:test`, the head integration suites, detekt on the four modules, and the ast-grep walls):

- **ce5fa63e fix(watchdog)** — the gate flake on `HeadServerCompactCapTest` (run 33575037270) was a real defect: `forCompact()` raised the first-output tier *to* the wall instead of turning it off, so two pollers shared one deadline and the verdict was a coin flip. The tier is now `Duration.INFINITE` for compactions; a new real-clock `WatchdogTest` arm pins it.
- **fb393038 fix(perf)** — the two malformed rows per perf file were an ENOSPC short write on 2026-08-25 (daemon-log: "No space left on device") that fused two rows into one unparsable line. `JsonlSink` now heals a torn tail before appending.
- **2c9e49d2 test(head)** — end-to-end proof of the capacity-signal handling this PR introduces: persistent 503 is retried by the head (POSTs counted) and surfaces as `overloaded_error`; a 503 that clears completes the turn; in-stream `server_is_overloaded` is `overloaded_error`, never `api_error`; a 403 wearing the code stays `invalid_request_error` with no retry.
- **eb73cd61 docs(launch)** — records why the non-streaming fallback stays enabled next to the `API_TIMEOUT_MS` contract.

Live proof of the recipe: launching the installed `claudex` and `claude-grok` wrappers against a stub `claude` on PATH shows `API_TIMEOUT_MS=960000` arriving in the child environment.

PR #116, opened 2026-09-02
test(e2e): fresh-machine Docker e2e (install → doctor → heads → real wrapper → uninstall)

marcospaulo:
## What

A fresh-machine end-to-end test that runs in Docker as if on a new machine, so a release can be proven installed and working before it is promoted.

`checks/e2e/docker/run.sh` builds an image with only the prerequisites (Java 21, Node 24, python3, curl, bash, Claude Code pinned to the host's version), mounts the artifacts and the checkout read-only, and runs `inside.sh` with `--network none`. Inside:

1. two mock upstreams on loopback: the migration oracle's vendored ChatGPT-backend mock (sliced from `dev/campaigns/proxy-hardening/oracle/mock-upstream.vendored.mjs`, not copied) for `openai-responses`, and a minimal `openai-chat` mock
2. a two-head topology (`claudex`, `claude-mockchat`)
3. `install.sh` from the artifacts (sha256sums verified → `init` → `install --all` → `doctor`)
4. `splice doctor`: every prerequisite ✓, no ✗ anywhere
5. cold start, `/health` ok with every head ready, `/api/heads`
6. `stream_probe.py` (the live e2e's contract oracle) and `count_tokens` through both heads
7. the launch recipe's contract per head: loopback `ANTHROPIC_BASE_URL`, `API_TIMEOUT_MS` above the daemon's 900 s wall
8. the daemon stopped, then a real `claudex -p` turn that makes the shim boot it, and a `claude-mockchat -p` turn
9. `splice restart`, `logs --tail`, `status`, `uninstall` (wrappers gone)

Every step lands in a JSON receipt with verdict, seconds and evidence; a failed step never stops the run and the exit code is non-zero on any failure. Receipts and the container's `daemon.log` go to `checks/e2e/receipts/` (gitignored).

```bash
npm run e2e:docker                                     # build this checkout, prove it
bash checks/e2e/docker/run.sh --release v0.3.0-beta.1  # prove a published release's assets
bash checks/e2e/docker/run.sh --jar P --shim P         # prove any prebuilt pair
```

## Evidence

| build | result | failing steps |
|---|---|---|
| v0.3.0-beta.1 release assets (jar `70f697a0…`) | 17/19 | launch recipe ×2: `API_TIMEOUT_MS` absent (the shipped defect #115 fixes) |
| #115 build (jar `f80a075d…`) | 19/19 | none |

Two harness traps worth knowing: `step()` captures its command in a command substitution, so the mocks report ports through files rather than shell variables; and Claude Code exports `CLAUDE_CODE_VERSION="X.Y.Z (Claude Code)"` into child shells, so the image's version arg is `SPLICE_E2E_CLAUDE_VERSION` (parsed to a bare version).

## Not in the merge gate

Needs Docker and ~1 min; proposed as the preflight before `npm run promote` (release ≠ promoted until this is green on the exact assets). Not wired into `checks/promote.sh` here.

### DIFF
--- a/checks/e2e/docker/Dockerfile
+++ b/checks/e2e/docker/Dockerfile
@@ -7,16 +7,24 @@
 # the wrapper move goes to the two in-container mock upstreams, or the run fails.
 FROM eclipse-temurin:21-jre-noble
 
-ARG CLAUDE_CODE_VERSION=latest
+# Every input is pinned so two builds a month apart prove the same machine (review of #116):
+# Node from the official tarball, verified against its published sha256; Claude Code at an exact
+# version (run.sh overrides it with the host's, when one is discoverable).
+ARG NODE_VERSION=24.20.0
+ARG NODE_SHA256=2f2c0da162318f0de47665410c7c8c2ed3d36c8f3105de4bbc61176c70a7cbf2
+ARG CLAUDE_CODE_VERSION=2.1.257
 # The unprivileged user's uid — run.sh passes the host's so the bind-mounted /out and /artifacts
 # are readable/writable on both sides (the noble base already owns uid 1000 as `ubuntu`).
 ARG UID=1000
 
-# iproute2 gives `ss` for the socket assertions; procps gives `pgrep` for the daemon assertions.
+# iproute2 gives `ss` for the socket assertions; procps gives `pgrep` for the daemon assertions;
+# xz-utils unpacks the Node tarball.
 RUN apt-get update \
- && apt-get install -y --no-install-recommends ca-certificates curl bash python3 iproute2 procps gnupg \
- && curl -fsSL https://deb.nodesource.com/setup_24.x | bash - \
- && apt-get install -y --no-install-recommends nodejs \
+ && apt-get install -y --no-install-recommends ca-certificates curl bash python3 iproute2 procps xz-utils \
+ && curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz" -o /tmp/node.tar.xz \
+ && echo "${NODE_SHA256}  /tmp/node.tar.xz" | sha256sum -c - \
+ && tar -xJf /tmp/node.tar.xz -C /usr/local --strip-components=1 --exclude='*/CHANGELOG.md' --exclude='*/README.md' --exclude='*/LICENSE' \
+ && rm /tmp/node.tar.xz \
  && npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
  && npm cache clean --force \
  && rm -rf /var/lib/apt/lists/*
--- a/checks/e2e/docker/Dockerfile
+++ b/checks/e2e/docker/Dockerfile
@@ -1,0 +1,30 @@
+# checks/e2e/docker/Dockerfile — a NEW MACHINE for the fresh-install e2e.
+#
+# Nothing splice-specific is baked in: the image is the prerequisite set the README asks a new
+# operator to have (Java 21+, Node 24, python3, curl, bash, Claude Code on PATH) and an unprivileged
+# user with an empty HOME. The release artifacts and the checkout are mounted at run time, so the
+# same image proves any build. The container runs with --network none: every byte the daemon and
+# the wrapper move goes to the two in-container mock upstreams, or the run fails.
+FROM eclipse-temurin:21-jre-noble
+
+ARG CLAUDE_CODE_VERSION=latest
+# The unprivileged user's uid — run.sh passes the host's so the bind-mounted /out and /artifacts
+# are readable/writable on both sides (the noble base already owns uid 1000 as `ubuntu`).
+ARG UID=1000
+
+# iproute2 gives `ss` for the socket assertions; procps gives `pgrep` for the daemon assertions.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl bash python3 iproute2 procps gnupg \
+ && curl -fsSL https://deb.nodesource.com/setup_24.x | bash - \
+ && apt-get install -y --no-install-recommends nodejs \
+ && npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
+ && npm cache clean --force \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN (userdel -r ubuntu 2>/dev/null || true) && useradd -m -u "${UID}" -s /bin/bash tester \
+ && mkdir -p /out && chown tester:tester /out
+
+USER tester
+WORKDIR /home/tester
+ENV HOME=/home/tester
+ENV PATH="/home/tester/.local/bin:${PATH}"
--- a/checks/e2e/docker/inside.sh
+++ b/checks/e2e/docker/inside.sh
@@ -19,6 +19,7 @@ OUT="${OUT:-/out}"
 CONTROL_PORT=3096
 CODEX_HEAD_PORT=3099
 CHAT_HEAD_PORT=3101
+CHAT2_HEAD_PORT=3105
 RECEIPT="$OUT/receipt.json"
 STEPS_FILE="$(mktemp)"
 FAILED=0
@@ -108,7 +109,7 @@ echo "fresh-machine e2e: user=$(id -un) home=$HOME artifacts=$ARTIFACTS"
 start_mocks() {
   nohup node "$REPO/checks/e2e/docker/mock_codex.mjs" "$REPO" 0 > "$OUT/mock_codex.out" 2> "$OUT/mock_codex.err" &
   echo $! > "$OUT/mock_codex.pid"
-  nohup python3 "$REPO/checks/e2e/docker/mock_chat.py" 0 > "$OUT/mock_chat.out" 2> "$OUT/mock_chat.err" &
+  MOCK_CHAT_HOLD_S=45 nohup python3 "$REPO/checks/e2e/docker/mock_chat.py" 0 > "$OUT/mock_chat.out" 2> "$OUT/mock_chat.err" &
   echo $! > "$OUT/mock_chat.pid"
   for _ in $(seq 1 50); do
     [ -s "$OUT/mock_codex.out" ] && [ -s "$OUT/mock_chat.out" ] && break
@@ -156,6 +157,18 @@ id = "mock-chat"
 label = "Chat (mock)"
 context_window = 128000
 
+# A second chat head on the SAME mock with its own window: the per-head contract check and the
+# cross-head ListAgents proof need two heads whose every turn the harness controls.
+[providers.mockchat2]
+dialect = "openai-chat"
+base_url = "http://127.0.0.1:$CHAT_MOCK_PORT"
+auth = { kind = "api-key", env = "MOCK_CHAT_API_KEY" }
+
+[[providers.mockchat2.models]]
+id = "mock-chat-2"
+label = "Chat 2 (mock)"
+context_window = 64000
+
 [heads.claudex]
 provider = "codex"
 port = $CODEX_HEAD_PORT
@@ -171,6 +184,14 @@ discovery_prefix = "claude-mockchat--"
 pinned_model = "mock-chat"
 [heads.mockchat.claude]
 command = "claude-mockchat"
+
+[heads.mockchat2]
+provider = "mockchat2"
+port = $CHAT2_HEAD_PORT
+discovery_prefix = "claude-mockchat2--"
+pinned_model = "mock-chat-2"
+[heads.mockchat2.claude]
+command = "claude-mockchat2"
 EOF
   cat "$HOME/.config/splice/splice.toml"
 }
@@ -205,6 +226,7 @@ install_step() {
   [ -x "$HOME/.local/bin/splice" ] || { echo "splice command not linked"; return 1; }
   [ -x "$HOME/.local/bin/claudex" ] || { echo "claudex wrapper not linked"; return 1; }
   [ -x "$HOME/.local/bin/claude-mockchat" ] || { echo "claude-mockchat wrapper not linked"; return 1; }
+  [ -x "$HOME/.local/bin/claude-mockchat2" ] || { echo "claude-mockchat2 wrapper not linked"; return 1; }
   ls -l "$HOME/.local/bin/"
   splice version
 }
@@ -226,7 +248,7 @@ step "doctor: prerequisites ✓, no ✗ anywhere" doctor_prereqs
 cold_start() {
   splice restart </dev/null || return 1
   wait_health 60 || return 1
-  ss -ltn | grep -E ":($CONTROL_PORT|$CODEX_HEAD_PORT|$CHAT_HEAD_PORT) " || { echo "head ports not listening"; return 1; }
+  ss -ltn | grep -E ":($CONTROL_PORT|$CODEX_HEAD_PORT|$CHAT_HEAD_PORT|$CHAT2_HEAD_PORT) " || { echo "head ports not listening"; return 1; }
 }
 step "daemon cold start: /health ok, every head ready" cold_start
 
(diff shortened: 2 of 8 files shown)


checks/e2e/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+ARG CLAUDE_CODE_VERSION=latest
+# The unprivileged user's uid — run.sh passes the host's so the bind-mounted /out and /artifacts
+# are readable/writable on both sides (the noble base already owns uid 1000 as `ubuntu`).
+ARG UID=1000
+
+# iproute2 gives `ss` for the socket assertions; procps gives `pgrep` for the daemon assertions.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl bash python3 iproute2 procps gnupg \
+ && curl -fsSL https://deb.nodesource.com/setup_24.x | bash - \
```
marcospaulo: **[concern]** The framing in the header is the right one, and the pinning was clearly thought about: the image is meant to be the fixed prerequisite set so "the same image proves any build", and run.sh spends three comment lines plus a parse on getting the Claude Code version pinned rather than floating. Two inputs in this RUN block still float, which erodes exactly that property.

**1. Node comes from an unpinned installer and a floating apt line.**

```dockerfile
 && curl -fsSL https://deb.nodesource.com/setup_24.x | bash - \
 && apt-get install -y --no-install-recommends nodejs \
```

Nothing pins a script revision or a nodejs version, so a nodesource-side change makes next month's image differ from this month's with no diff in the repo. For a receipt whose value is "v0.3.0-beta.1 scored 17/19, the #115 build scored 19/19", an environment that can move between those two runs is a confound in the comparison. Pinning the package is the cheap half of the fix, verifying the script is the thorough half:

```dockerfile
 && curl -fsSL https://deb.nodesource.com/setup_24.x -o /tmp/nodesource.sh \
 && echo "<sha256>  /tmp/nodesource.sh" | sha256sum -c - \
 && bash /tmp/nodesource.sh \
 && apt-get install -y --no-install-recommends nodejs=24.* \
```

**2. `ARG CLAUDE_CODE_VERSION=latest` (line 10) feeds line 20, and through run.sh that default turns out to be unreachable in a way worth knowing about.** run.sh:64 is:

```bash
CC_VERSION="$(printf '%s' "${SPLICE_E2E_CLAUDE_VERSION:-$(claude --version 2>/dev/null)}" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)"
```

Under `set -euo pipefail` (run.sh:17), a grep that matches nothing exits 1, pipefail promotes that to the pipeline's status, and the assignment inherits it. I ran that exact shape to check: the `${CC_VERSION:-latest}` on line 65 is never reached and the shell exits 1 right there. So on a host without `claude` on PATH and without `SPLICE_E2E_CLAUDE_VERSION` set (a CI runner, a fresh clone), `npm run e2e:docker` dies silently before the build even starts, and the `latest` default is only reachable by hand-building the image and then using `run.sh --no-build`.

Two small changes cover both halves:

```bash
# run.sh:64 - let the documented fallback actually happen, and say so
CC_VERSION="$(printf '%s' "${SPLICE_E2E_CLAUDE_VERSION:-$(claude --version 2>/dev/null)}" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || true)"
[ -n "$CC_VERSION" ] || echo "run.sh: no claude version discovered, building with the Dockerfile default" >&2
```

plus giving `ARG CLAUDE_CODE_VERSION` a real version as its default instead of `latest`, so a hand-built image is pinned too and the README's "the image pins Claude Code" line stays true on every path.

Separating the two: the nodesource pin is a suggestion (the risk is slow drift), the run.sh:64 abort is a defect on the documented default path and worth fixing regardless of what you decide about pinning.

checks/e2e/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ && apt-get install -y --no-install-recommends nodejs \
+ && npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
+ && npm cache clean --force \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN (userdel -r ubuntu 2>/dev/null || true) && useradd -m -u "${UID}" -s /bin/bash tester \
```
marcospaulo, line 32: **[concern]** The uid plumbing here is deliberate and mostly right. Passing the host uid through `--build-arg "UID=$(id -u)"` (run.sh:66) and clearing the noble base's uid-1000 `ubuntu` first is exactly what makes the `/out` bind mount writable from both sides.

One host shape breaks it. When run.sh itself runs as root (a root shell, `sudo bash run.sh`, a uid-0 CI job container), `$(id -u)` is `0` and this line becomes `useradd -m -u 0 -s /bin/bash tester`:

```dockerfile
RUN (userdel -r ubuntu 2>/dev/null || true) && useradd -m -u "${UID}" -s /bin/bash tester \
```

`useradd(8)` documents `-u` as "this value must be unique, unless the -o option is used", with exit value 4 = "UID already in use (and no -o)". uid 0 already belongs to root on this base, so the RUN fails and the whole `docker build` aborts. It surfaces as a bare useradd exit 4 in build output, which reads like an image bug rather than "you are root".

`-o` would make it build, but then `USER tester` is uid 0 and the image quietly stops being the unprivileged empty HOME the header promises. Failing fast in run.sh keeps that property and says why:

```bash
HOST_UID="$(id -u)"
[ "$HOST_UID" != 0 ] || { echo "run.sh: run as a non-root user (the image needs a non-zero uid for the unprivileged tester account)" >&2; exit 2; }
docker build -q --build-arg "CLAUDE_CODE_VERSION=${CC_VERSION:-latest}" --build-arg "UID=$HOST_UID" \
  -t "$IMAGE" "$ROOT/checks/e2e/docker" >/dev/null
```

Falsifiable either way: `docker build --build-arg UID=0 checks/e2e/docker` should reproduce the exit 4.

checks/e2e/docker/inside.sh
```diff
@@ -0,0 +1,318 @@
...
+CODEX_MOCK_REFRESH="http://127.0.0.1:$CODEX_MOCK_PORT/oauth/token"
+export MOCK_CHAT_API_KEY="mock-chat-key"
+export CODEX_OAUTH_TOKEN_URL="$CODEX_MOCK_REFRESH"
+
+# ── 3. install from the artifacts, exactly as a release install verifies them ─────────────────
+install_step() {
+  [ -f "$ARTIFACTS/splice.jar" ] || { echo "no $ARTIFACTS/splice.jar"; return 1; }
+  [ -f "$ARTIFACTS/splice-launch" ] || { echo "no $ARTIFACTS/splice-launch"; return 1; }
+  if [ -f "$ARTIFACTS/sha256sums.txt" ]; then
+    (cd "$ARTIFACTS" && sha256sum -c sha256sums.txt --ignore-missing) || return 1
```
marcospaulo: **[concern]** Verifying the release manifest before installing is the right shape, and the else branch is honest about the checkout path where no `sha256sums.txt` exists. The gap is that this check takes its denominator from the file it is checking, so it cannot fail for an artifact the manifest never mentioned:

```bash
  if [ -f "$ARTIFACTS/sha256sums.txt" ]; then
    (cd "$ARTIFACTS" && sha256sum -c sha256sums.txt --ignore-missing) || return 1
```

I reproduced it before writing this. With `splice.jar` and `splice-launch` both present on disk but absent from the manifest, `sha256sum -c sums.txt --ignore-missing` prints only `install.sh: OK` and exits 0. Line 197 then installs that jar and shim, and the step named "install.sh from artifacts: jar+shim verified" records PASS having hashed neither of them.

The total-omission case does fail closed (coreutils prints "no file was verified" and exits 1, also tested), so the live hole is the partial manifest: a release workflow that covers some assets and drops one. That is the exact shape a manifest bug takes, and it is the one this step is positioned to catch.

Asserting the two names the step claims to verify keeps the receipt honest:

```bash
  if [ -f "$ARTIFACTS/sha256sums.txt" ]; then
    local sums f
    sums="$(cd "$ARTIFACTS" && sha256sum -c sha256sums.txt --ignore-missing 2>&1)" || { printf '%s\n' "$sums"; return 1; }
    printf '%s\n' "$sums"
    for f in splice.jar splice-launch; do
      printf '%s\n' "$sums" | grep -qx "$f: OK" || { echo "$f is not covered by sha256sums.txt"; return 1; }
    done
```

Reading this as a requirement rather than a preference, because the step name is the claim the receipt carries downstream.

checks/e2e/docker/inside.sh
```diff
@@ -0,0 +1,318 @@
...
+step "wrapper turn: claudex -p boots the daemon and completes" wrapper_turn claudex "ok after auth"
+step "wrapper turn: claude-mockchat -p through the head" wrapper_turn claude-mockchat "END"
+
+# ── 9. restart, logs, status, uninstall ────────────────────────────────────────────────────────
+restart_step() {
+  splice restart </dev/null || return 1
+  wait_health 60
+}
+step "splice restart: healthy again" restart_step
+step "splice logs --tail 5" bash -c 'splice logs --tail 5 </dev/null'
```
marcospaulo: **[concern]** The fail-closed discipline through the rest of this file is genuinely strong: `doctor_prereqs` greps for each `✓` and for the absence of any `✗`, `api_heads` asserts the exact head keys, `launch_recipe` asserts the base URL and that `API_TIMEOUT_MS` clears the 900s wall, `wrapper_turn` greps for the mock's own reply. These two steps are the only ones in the file that assert nothing:

```bash
step "splice logs --tail 5" bash -c 'splice logs --tail 5 </dev/null'
step "splice status" bash -c 'splice status </dev/null'
```

`step()` derives the verdict from the exit status alone (inside.sh:52-58) and nothing downstream reads `$out`, so both record PASS as long as the process exits 0. That is weaker than the step names in a specific, testable way: a `status` that prints "daemon: not running" and exits 0 is a normal CLI convention (the command succeeded at reporting, the subject was unhealthy), and an empty log tail exits 0 too. The receipt would carry `"splice status": "PASS"` for a run where status reported a dead daemon. Since both run immediately after `restart_step` has already waited for health, as written they add close to zero information to the receipt.

Matching the assertion style used everywhere else in the file closes it:

```bash
logs_step() {
  local out
  out="$(splice logs --tail 5 </dev/null 2>&1)" || return 1
  printf '%s\n' "$out"
  [ -n "$out" ] || { echo "empty log tail"; return 1; }
}
status_step() {
  local out
  out="$(splice status </dev/null 2>&1 | strip_ansi)" || return 1
  printf '%s\n' "$out"
  printf '%s\n' "$out" | grep -q 'claudex' || { echo "status does not mention the claudex head"; return 1; }
}
step "splice logs --tail 5" logs_step
step "splice status" status_step
```

I do not know the exact wording `splice status` prints, so treat that grep target as a placeholder for whichever marker you consider load-bearing (the head names, "running", the control port).

checks/e2e/docker/run.sh
```diff
@@ -0,0 +1,86 @@
...
+  -v "$ART:/artifacts:ro" -v "$ROOT:/repo:ro" -v "$RUN_OUT:/out" \
+  "$IMAGE" bash /repo/checks/e2e/docker/inside.sh
+RC=$?
+set -e
+if [ -f "$RUN_OUT/receipt.json" ]; then
+  cp "$RUN_OUT/receipt.json" "$OUT/docker-$STAMP.json"
+  mkdir -p "$OUT/docker-$STAMP" && cp -r "$RUN_OUT"/. "$OUT/docker-$STAMP/"
+  echo "run.sh: receipt $OUT/docker-$STAMP.json (steps + daemon.log in $OUT/docker-$STAMP/)"
+fi
+rm -rf "$RUN_OUT" 2>/dev/null || true
```
marcospaulo: **[suggestion]** `ART` is handled exactly right: created with `mktemp -d`, then `trap cleanup EXIT` on line 40, so it goes away on every exit path including a mid-script failure. `RUN_OUT` is the same kind of temp dir but is not covered by that trap:

```bash
rm -rf "$RUN_OUT" 2>/dev/null || true
```

This line runs after `set -e` is restored on line 79, with three commands able to fail in between: the `cp` of receipt.json, the `mkdir -p`, and the `cp -r` into `$OUT/docker-$STAMP/`. If the receipts dir is not writable, the disk fills, or a stale cached image built with a different `--build-arg UID` writes files the host user cannot read, `set -e` aborts before line 85 and it never runs. What survives is a mode 0777 directory in `$TMPDIR` (chmod on line 72) holding daemon.log, the mock stdout and stderr, and every step log, one per failed invocation.

Extending the trap you already have covers it and keeps `--keep` meaning one thing:

```bash
RUN_OUT=""
cleanup() {
  if [ "$KEEP" = 1 ]; then
    echo "artifacts kept at $ART${RUN_OUT:+, run output at $RUN_OUT}"
  else
    rm -rf "$ART"
    [ -z "$RUN_OUT" ] || rm -rf "$RUN_OUT"
  fi
}
trap cleanup EXIT
```

with `RUN_OUT` still assigned where it is today on line 71 (declaring it empty first keeps `set -u` happy if the trap fires before then).

Preference rather than requirement: this only fires on an error path and costs a stale temp tree, no wrong result and no false green. Worth closing mainly because the correct pattern is already in the file two dozen lines up.

marcospaulo (commented):
Review of #116: a Docker fresh-machine end-to-end harness (install from artifacts, doctor, cold start, both dialect heads at the wire, launch recipe contract, real Claude Code wrapper, restart, logs, status, uninstall) with a per-step receipt.

Findings: 4 concerns, 1 suggestion. Dockerfile: useradd aborts the image build when the host uid is 0 (no -o), and the nodesource script plus the nodejs package float while the file's own header promises a reproducible image. inside.sh: the logs and status steps grade on exit code alone, so an empty or unhealthy report still passes; the checksum step's --ignore-missing makes its coverage self-defined (reproduced: a manifest that omits splice.jar still exits 0). run.sh: RUN_OUT is the one mktemp dir not covered by the EXIT trap, and its rm sits behind three failure-capable commands after set -e returns. 4 raw findings dissolved by the evidence gates (curl-pipe-bash as a generic prior, folded into the pinning concern; the token-URL indirection, which carries no credential; unredacted logs in a gitignored receipts dir with only per-container throwaway secrets in scope; one duplicate consolidated).

Checked clean: the npm script resolves to an added file and follows the e2e:* naming; the mock ports and pids cross the step() subshell boundary through files; every other step in inside.sh asserts content, not just exit status; the receipt fails closed on any step; no credential-shaped literal remains for the CI secret pass.

Mode: fan-out (609 added lines), three lenses (correctness deep on the five harness files, security and build-config on skim digests). Block 3b flow map skipped: pure test tooling. Mechanical scan: run (94 rules, all Kotlin-oriented, so shell, python and Dockerfile coverage from the packs is nil). Mechanical-only coverage: checks/e2e/README.md. No reduced-confidence findings. No antibody drafts offered.

checks/e2e/docker/Dockerfile
marcospaulo: Fixed in b44e0f9d: Node now comes from the official tarball pinned at 24.20.0 with its sha256 verified in the Dockerfile (no nodesource script, no floating apt package), CLAUDE_CODE_VERSION defaults to an exact version, and run.sh's version-discovery pipeline no longer aborts under pipefail: an empty result announces the fallback and builds with the Dockerfile pin.

checks/e2e/docker/Dockerfile
marcospaulo: Fixed in b44e0f9d: run.sh refuses uid 0 with the reason before docker build. Reproduced first: `docker build --build-arg UID=0` fails on useradd as described.

checks/e2e/docker/inside.sh
marcospaulo: Fixed in b44e0f9d: the checksum step asserts `splice.jar: OK` and `splice-launch: OK` by name. Mutation-tested: a manifest that omits them fails, a corrupted jar fails, the complete manifest passes.

checks/e2e/docker/inside.sh
marcospaulo: Fixed in b44e0f9d: logs_step requires a non-empty tail; status_step requires a `daemon running` row plus both head rows (claudex, claude-mockchat). Fresh-machine run on the current #115 build: 19/19.

checks/e2e/docker/run.sh
marcospaulo: Fixed in b44e0f9d: RUN_OUT is declared empty up front and removed by the same EXIT trap as ART; --keep now reports both paths.

PR #120, opened 2026-09-02
feat(statusline): every head draws its plan usage, and the line follows the picked row

marcospaulo:
Stacked on #116 (base is its branch); merge after it.

## The status line follows the picked row (1822aa22)

Claude Code fixes its context window per process (the pinned row's) and splice scales the counts it reports so any other row compacts at its own window. That left the daemon's status line showing the session's window and the scaled counts however the operator switched: picking `grok-4.6[500k]` on a 256k grok head still read `…/256k`.

`ManagedHead` carries the head's `ModelCatalog`; the renderer looks the picked id up in it through the new `StatuslineRow` lens. A declared row renders by its label, its declared window, and its counts divided by the same factor splice multiplied them by. The pinned row and a catalog-less head render the blob exactly as before. `StatuslineScaledRowTest` pins both.

## Every head draws its plan usage (19dd9127)

The operator's status line on the native Claude head shows `model·effort  $spend  5h ████░░░░ 14%  7d ███░░░░░ 42%` with the reset time once a bar matters. Every splice head now draws the same line from its own provider's quota:

- **Sources, per head.** Every upstream response is read for Anthropic's unified rate-limit headers (passthrough) and the `x-codex-*` round headers; chatgpt-oauth, kimi-oauth and grok-oauth heads also poll their provider's usage endpoint every five minutes (ChatGPT `backend-api/wham/usage`, Kimi `v1/usages`, SuperGrok `v1/billing`), shapes captured live on 2026-09-02 and pinned as fixtures in `QuotaParsersTest`. Windows are sorted into the two slots by their own length (≤ 6h is the 5h slot): Codex Pro calls its weekly window "primary" and has no 5-hour one.
- **The client reads the same numbers.** The tracked windows ride every head response as `anthropic-ratelimit-unified-{5h,7d}-utilization/-reset`, which Claude Code reads into the `rate_limits` it hands the status line. Before the first response of a session the daemon's own snapshot fills the bars. `/usage/<head>` carries the snapshot as `quota`; state lives in `<head>-quota.json`.
[...]

PR #121, opened 2026-09-02
fix(ws): a close explains itself, a torn compaction restarts, and a big frame is not a stall

marcospaulo:
Stacked on #120 (base `fix/statusline-scaled-row`), merges after it.

## The backend is not the story

I first read these socket closes as the backend dropping streams. That was wrong, and the operator called it: "they dont fail this much as normal operations". Re-read across the whole four-day log rather than the bad hour, 48,898 turns:

| failed turns | 737 | 1.51% |
|---|---|---|
| our own watchdog stalls | 147 | the byte-tier bug, already fixed on this lineage |
| backend policy verdicts | 270 | `cyber_policy` and refusals, deterministic and correct |
| backend said overloaded | 303 | **302 of them inside a single incident day**; 0, 1, 0 on the other three |
| upstream stream truncated, surfaced to the client | 17 | the tears the re-anchor could not absorb |

That last row is the failures the client saw, not the number of tears. Counting absorbed ones too, streams tore 79 times in four days, between 0.06% and 0.24% of turns per day. Roughly twenty a day, and about four fifths of them are invisible because the re-anchor restarts the round and the turn still succeeds. That is not a backend that fails a lot, and it is not nothing either. What made it look like an outage was our own log line.

## The line was making an accusation the client cannot support

`status=1006` is **reserved** by RFC 6455 §7.4.1 and can never be sent by a peer. Our JDK client synthesises it whenever the TCP stream ends with no close frame, in `WebSocketImpl`'s receive task via `onComplete → CLOSED_ABNORMALLY`. From our side the origin, a load balancer, a proxy hop and the network are indistinguishable. "socket closed by the server" was therefore unearned every single time it was printed.

I also tested the opposite hypothesis, that the line was echoing our own `kill()`. It is not: `abort()` calls `receiveScheduler.stop()` before `close()`, so an aborted socket never delivers `onClose` to our listener. These are real inbound stream ends. They are just not evidence of anyone's failure.

Read with the new clauses, today's twelve end-of-stream lines say something ordinary. Six followed between 18 and 1,483 seconds of idle time on sockets our pool never ages out, which any infrastructure reaps. Eleven of the twelve ended no round at all. One coincided with a compaction failure, and the socket id that would prove it was even the same socket is precisely what was missing.

## Changes

- **The end-of-stream line reports the observation, not a culprit.** 1006 now reads "socket stream ended with no close frame (status=1006, actor unknown from here)". A real close frame still names the peer, its code and its reason, because that one is attributable. Beside it: socket label, age, idle or mid-round with elapsed, last frame, last peer ping, open socket count. `onPing` is recorded and re-arms demand.
- **Compaction re-anchors like every other round.** It was the one round shape excluded, on the belief that the pre-stream retry covered it, and it did not: a stream ending without `response.completed` reaches the translator, not the pre-handoff interlock. A compaction's partial is usage-only, so its only continuation is the verbatim whole request with backoff, which is the retry Codex CLI performs for its remote compaction. Deterministic verdicts still carry no partial and are never re-POSTed. This matters because tears run around twenty a day and a compaction is a 244k-token read, so the one round shape that could not absorb a tear was also the most expensive one to lose.
- **The send budget scales with the frame.** It was a flat 10s sized when the largest frame was 1.5 MB. Frames now reach 7.7 MB, and all 13 of one morning's stalls landed while a 5 to 6.5 MB frame was in flight. Healthy sockets poisoned for being asked to swallow megabytes, then the same megabytes re-sent over SSE. Now the floor plus transfer time at 100 KB/s, and the line names both the budget and the frame size.

## Verified

- `npm run gate:rules` 47 passed; concentration ratchet holds at the HIGH baseline of 0.
- `:dialect-openai-responses:test` and `detekt` green, 364 tests. New: the wording wall proving 1006 names no actor while a close frame is attributed to the peer, `WsPulseTest` ×3, close-diagnostics ×3, the frame-scaled send budget, the compact re-anchor seam, and the controller's verbatim restart.
- `:gateway:test` and `detekt` green in the 4096-task box.
- Docker fresh-machine e2e 30/30 on the jar built from this branch merged with #115, receipt in the comments.

Not done here, on purpose: our pool has no max age and no idle eviction, which is why we hold sockets for 25 minutes and then discover the far side reaped them. It costs one reconnect and no failure, so it is not urgent, and the new clauses are what would justify picking a number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Added
 
+- **Every head shows its plan usage the way the native Claude head does.** The daemon tracks each
+  head's 5-hour and 7-day windows (ChatGPT's usage endpoint and its `x-codex-*` round headers,
+  Kimi's usage endpoint, SuperGrok's billing period, Anthropic's own unified headers on the
+  passthrough head), stamps them onto every response as the `anthropic-ratelimit-unified-*`
+  headers Claude Code reads into its `rate_limits`, and draws them on the status line as the 5h
+  and 7d bars with the reset time once a bar is worth acting on, beside effort and session spend.
 - **`claude-splice`, the native-auth Claude head.** Claude Code keeps its own Anthropic login and
   sends the caller credential through the local passthrough head; splice never stores, reads,
   refreshes, or logs that credential. The management key is not reused on this route.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **Compactions no longer die on the ChatGPT backend's capacity signal.** An in-stream
+  `server_is_overloaded` (or `slow_down`) used to classify as a non-retryable `api_error`, so the
+  proxy neither reissued nor salvaged the turn and Claude Code reported the compaction failed. Any
+  overload-shaped error code on an in-stream or server-side (5xx) failure is now a transient
+  `overloaded_error`, the type Claude Code retries with backoff, matching codex's own handling of
+  the same two codes; a 4xx keeps its deterministic verdict whatever its code spells.
+- **Claude Code now waits out the proxy's own whole-turn wall.** Every wrapper plants
+  `API_TIMEOUT_MS` from the head's `upstreamTimeoutMs` plus a minute of grace (960 s on the
+  default 900 s cap). With Claude Code's 600 s default, every compaction longer than ten minutes
+  ended as a client abort while the daemon was still streaming the summary.
+- **A silent compaction is ended by the whole-turn wall alone.** The compact budget used to raise
+  the first-output tier to the wall instead of switching it off, leaving two pollers on one
+  deadline; on a loaded runner the idle poller could win the tick and end a round as "first-output
+  cap" (salvage invited) where the wall should have ended the turn. The tier is now off for
+  compactions; `streamIdle` still reaps a stall once output has begun.
+- **A torn perf row no longer swallows the row after it.** When the disk filled mid-append
+  (2026-08-25) a short write left a fragment with no newline and the next row fused onto it, so
+  readers lost both. The JSONL sink now heals a torn tail before appending.
+
 ## splice v0.3.0-beta.1 — native Claude auth and a hardened multi-head gateway - 2026-08-30
 
 ### Added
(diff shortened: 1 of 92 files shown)


marcospaulo:
Docker fresh-machine e2e on this branch merged with #115 (the API_TIMEOUT_MS launch recipe the harness asserts): **PASS 30/30**.

```
jar    7eca24e4c1790444…  (this branch @ 1a8d55b8 + origin/fix/compaction-overload-and-client-timeout)
shim   29f7070d9ad1bb04…
receipt docker-20260902T162947Z.json
```

Every step green, including the claudex wire probe over the mock (openai-responses), the wrapper turn that boots the daemon, and plan usage on the status line.

marcospaulo:
Correction, and why the PR description changed.

My first pass here read the 1006 closes as the backend dropping streams. Re-reading the whole four-day log instead of the bad hour, that is not what the data says. Of 737 failed turns in 48,898, the backend accounts for 17 torn streams (0.035%) plus 303 overload verdicts of which 302 fall inside one incident day. Our own watchdog stalls account for 147 and the backend's policy classifier for 270.

The wording of our own log line is what made the wrong reading easy. RFC 6455 reserves 1006, so no peer can send it. Our JDK client synthesises it whenever the stream ends without a close frame, which makes the origin, a load balancer and the network indistinguishable from here. `d6da38e6` changes the line to say the stream ended with the actor unknown, and keeps attribution only for a real close frame, where the peer's own code and reason are on the wire.

I checked the opposite hypothesis before landing it, that the line was echoing our own `kill()`. It is not. `abort()` stops the receive scheduler before closing, so an aborted socket never delivers `onClose` to the listener.

There is a test that fails if the word comes back: `1006 names no actor, while a real close frame is attributed to the peer`.

marcospaulo:
Second correction, to my own figure above. The 17 truncated turns are the tears that SURFACED, not the tears. Counting the ones the re-anchor absorbed, streams tore 79 times across the four days, 0.06% to 0.24% of turns per day, roughly twenty a day with about four fifths invisible. Still not an outage, and still mostly our own caps and pooling in the failure column, but the tear rate is an order of magnitude above what I first wrote. It also sharpens why compaction belongs in the re-anchor: tears are routine enough that excluding the most expensive round shape from the absorber was the actual defect.

PR #126, opened 2026-09-04
chore(release): v0.3.1 — one picker row per model, and the proofs the review asked for

marcospaulo:
## What

A patch on top of v0.3.0 (which nobody has installed yet) carrying what the 2026-09-04 release demo found: a Codex session reviewing a commit for a Claude session, and the `/model` picker on camera.

- **fix(launch): a model that fills a slot is not also a picker row** (1094b0cc). Every head listed its slotted models twice (Sol twice on claudex, K3 (256k) and K2.7 Code twice on claude-kimi). Claude Code 2.1.257 renders `ANTHROPIC_DEFAULT_*_MODEL` as tier rows valued `opus`/`sonnet`/... with the `_NAME` label, then appends `additionalModelOptionsCache` rows valued by the raw id, and dedupes on value only. The materialized cache now omits ids already planted in a slot; `availableModels` is untouched.
- **test(watchdog)** (c46ba3ff). `TurnLineWatchdogVerdictTest` fabricated a compact first-output fire at a "lifted" 900s tier that production never produces (`forCompact` switches the tier off); it now pins a non-compact first-output fire and a compact total-cap fire. Two `TurnFinishTest` arms fire real pollers on virtual ticks and assert the sentinel reaches the rendered line through the production hop. `ManagedHeadFactory` gains a poller-start seam and `ManagedHeadFactoryQuotaPollTest` proves `CLAUDEX_QUOTA_POLL=off` starts nothing. `accept.sh` exercises the stable pin recipe. The "Heads that see each other" bullet moves to the v0.3.0 section, where it landed.
- **chore(release): v0.3.1** (4560ed51). Four version sites, CHANGELOG section. SHIM_VERSION stays shim-2.

## Evidence

- `:control:test --tests '*LaunchService*'`: 18/18, the new arm fails on the previous LaunchService.
- `:gateway:test` TurnFinishTest + TurnLineWatchdogVerdictTest, `:app:test` ManagedHeadFactoryQuotaPollTest, detekt: green; each new test fails under the named mutation (swapped sentinels, inverted guard).
- `stage.sh` + `accept.sh`: `release accept: OK`.
- Full local gate is running in the buildgate box; result to follow in a comment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #127, opened 2026-09-04
test(e2e): the head contract reads the picker as slots or cache rows, never both

marcospaulo:
The Docker fresh-machine e2e on main 9a49420b failed 27/30: its head contract step asserted an `additionalModelOptionsCache` row per catalog model, which v0.3.1's picker dedupe (#126) deliberately no longer materializes for a slotted model. The step now reads the slot set from the recipe's `ANTHROPIC_DEFAULT_*_MODEL` values and asserts the new rule: never both, slots ∪ rows = roster, cache rows keep their window.

Evidence: a fresh `npm run e2e:docker` on this branch, result to follow in a comment.


marcospaulo:
`npm run e2e:docker` on this branch: `FRESH-MACHINE E2E: PASS (30/30 steps)` (log in the buildgate box, receipts under checks/e2e/receipts/). On main 9a49420b the same run was 27/30 with the three head-contract steps failing on an empty cache.

PR #129, opened 2026-09-05
fix(launch): a tier that repeats another tier's model is planted under the wrapped id, so the picker draws it once

marcospaulo:
The 0.3.1 dedupe (#126) removed slotted models from additionalModelOptionsCache and did not
remove the duplicate: re-recorded on the fixed daemon, /model on claudex still listed Sol twice
and on claude-kimi K3 and K2.7 Code twice, and the Default line lost its label ("currently
gpt-5.6-sol[1m]"). Read against Claude Code 2.1.257's picker: it draws one row per PLANTED tier
(fen/hen/uen emit a row whenever ANTHROPIC_DEFAULT_<tier>_MODEL is set) and dedupes rows by
alias, so Fable sharing the frontier with Opus, and Haiku sharing Sonnet on a two-model head, is
the duplication; cache rows that repeat a tier's model were already dropped by Claude Code, and
they are where the Default line's label comes from.

A tier cannot be left unset: the alias then resolves to Claude Code's built-in model, which the
head rejects ("proxies its own models only"), and every fable- or haiku-tiered subagent dies.
The lever is the allowlist. A repeated tier is now planted as "<discovery prefix><id>": the head
routes it like the bare id (catalog.contains unwraps; count_tokens on the claudex head answered
200 to claude-codex--gpt-5.6-sol), and the picker hides a tier whose model is not on
availableModels. Verified live 2026-09-04 with the recipe env edited by hand: eight rows, Sol
once, "currently Codex 5.6 Sol". The cache is materialized in full again.

LaunchSpec carries the head's discoveryPrefix (blank keeps the bare duplicate). LaunchServiceTest
pins the wrapped haiku and fable on a positional two-model head and on the codex roster, and that
a prefix-less head keeps every cache row; 20 tests. The Docker e2e head contract goes back to a
cache row per model and adds: the four tier values are distinct and point inside the roster.
CHANGELOG 0.3.1 says what the fix is.

Evidence to follow in comments: local gate, Docker e2e on this branch, and the /model recording on this machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,15 @@
 
 ### Fixed
 
-- **The `/model` picker lists each model once.** Every head showed its slotted models twice (Sol
-  twice on `claudex`; Kimi K3 (256k) and Kimi K2.7 Code twice on `claude-kimi`). Both rows were
-  ours: the launch recipe plants `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU,FABLE}_MODEL`, which Claude
-  Code renders as its tier rows with the `_NAME` label, and then appends every
-  `additionalModelOptionsCache` row under the raw catalog id, and its dedupe compares values only.
-  The materialized cache now omits any id already planted in a slot; `availableModels` keeps every
-  id, so a saved model choice stays allowed, and unslotted rows keep their fields.
+- **The `/model` picker lists each model once.** Every head showed a slotted model twice (Sol
+  twice on `claudex`; Kimi K3 (256k) and Kimi K2.7 Code twice on `claude-kimi`). Claude Code draws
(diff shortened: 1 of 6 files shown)


marcospaulo:
Docker fresh-machine e2e on this branch tip (1a720f1c):

```
run.sh: artifacts: 8768e339d75b3657… splice.jar, 9f48bd9a0558eabf… splice-launch
FRESH-MACHINE E2E: PASS (30/30 steps)
```

The head contract now asserts one picker row per planted tier, bare tier ids distinct, wrapped repeats ending in the roster id. Local gate (`checks/gate.sh` in the buildgate box) passed on the same tree. The daemon on this machine runs this build and `/model` on claudex and claude-kimi was re-recorded with each model listed once and the Default row labelled.

PR #137, opened 2026-09-06
feat(head): every OAuth head owns its credential, a compaction outlives its client, and the activity side query is answered locally

marcospaulo:
## Why

Three things a day of Codex sessions on splice showed (2026-09-05), plus one credential rule the operator set.

- **Compactions cut off by the client.** Claude Code abandons an auto-compaction at 600 s (a hard client constant, not `API_TIMEOUT_MS`) and retries the same bytes minutes later. Astra compactions run 5-10 minutes, so 7 of the day's 78 were cancelled upstream at 600 s and re-run cold.
- **A side query burning tokens.** Every 30 s during a subagent turn the client re-sends the whole transcript asking for a 3-5 word present-tense activity label: 294M input tokens in a day at 48% cache hit, for text nobody reads twice.
- **Cold reconnects.** Turns after a WebSocket reconnect read the prompt cache cold; codex-rs sends session/thread/routing headers splice did not.
- **Shared credentials.** The codex head read `~/.codex/auth.json`, the same file codex-cli and the ChatGPT desktop app refresh. An OpenAI-style refresh rotates the refresh token, so each side signed the other out; one rotation left the head with "no upstream credentials" for 16 s (24 failed turns). Operator law: native apps and splice heads keep separate credentials, and may be separate accounts.

## What

- **A compaction outlives its client.** A compact stream turn drives on a scope the call's cancellation cannot reach and records its frames (`FrameRecording`); a lost client detaches the channel, the turn finishes, and the byte-identical retry is served from the recording or follows the live turn (`CompactionReplay`, keyed on session + body hash, 2 h TTL). Slot and recording are settled by the drive's own `finally`. A head stop ends compactions still driving; the scope survives a restart (the first cut cancelled it: the first compaction after a stop-start came back empty with a leaked slot, pinned by a regression test).
- **The activity-label side query is answered locally** (`ActivityLabel`, `LocalResponses`): recognised by its prompt, answered from the transcript's last tool call, no upstream turn.
- **codex-rs headers ride every turn**: `session-id`, `thread-id`, `x-codex-routing-hint`, and `x-client-request-id` on the WebSocket handshake (`CodexRoutingHeaders`, a `perTurnHeaders` hook on the Responses provider; grok's conv id moves onto the same hook).
- **Learned client windows persist** per head (`<head>-client-windows.json`), so the first turn after a restart is scaled right. A rejected request body logs its byte counts and failure class.
- **Every OAuth head signs in on its own credential file.** `chatgpt-oauth`, `grok-oauth`, `kimi-oauth` default to `~/.config/splice/auth/{codex,grok,kimi}.json`, written by `splice login <head>`; the apps' files are used only when `auth.file` names one. The auth-kind registry is the single source (splice-owned `authFile` plus the native app's file per kind); the `CODEX_AUTH_PATH` / `GROK_AUTH_PATH` knobs reference it; the login command and the kimi and openai-chat arms resolve through it. `splice doctor` warns on a config that still names an app's file, with the exact fix. README, example config and CHANGELOG updated.

## Verified

- Standalone tree (HEAD + this diff): 1201 module tests, 0 failed; detekt and the ast-grep walls clean; oracle replay 11/11 byte-match; CX-19 wall green; ledger floor OK.
- Red-green: the stop-start compaction regression test failed on the old scope handling and passes now; the registry-default test fails on any native-file default.
- Live 2026-09-05: side-query answered locally within minutes of install (0 upstream rows for it since); persisted windows restored after restart; a compaction cut by a restart continued detached and was held for its retry; all three heads signed in on splice-owned files, doctor clean.

Stacked on #136 (base `fix/compaction-is-a-turn`). The code-mode work in the same checkout is deliberately not in this PR and comes separately.

### DIFF
--- a/gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
+++ b/gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
@@ -1,0 +1,77 @@
+// NEW: the answers of compactions whose clients gave up, held for their retries (2026-09-05).
+//
+// Claude Code arms a 600 s wall-clock cap on an auto-compaction (`dKt=600000`, `recovery-timeout`,
+// not reset by frames), aborts at it, and retries the SAME request minutes later — on 2026-09-05
+// eight claudex compactions died at 600-602 s and four of the six retries were byte-identical, each
+// abort throwing away a 600 s upstream read. So a compaction's turn outlives its client
+// (TurnStreamer.driveDetachable), its frames are recorded (FrameRecording), and a retry that would
+// send the SAME BYTES upstream is answered from the recording (TurnPreparation → LocalResponses):
+// complete, in one burst, or still in flight, followed to its end. Keyed on session + a hash of the
+// upstream request body, which is what "the same compaction" means on the wire — a retry that
+// differs (a message arrived in between) simply runs upstream as before. Only compactions whose
+// client actually left are kept; a compaction delivered to its client has no retry to serve.
+package splice.gateway.head
+
+import splice.core.util.ElapsedClock
+import splice.core.util.MonoClock
+import splice.gateway.wire.FrameRecording
+import java.security.MessageDigest
+
+internal class CompactionReplay(
+    private val clock: ElapsedClock = ElapsedClock(MonoClock::nowMs),
+    private val ttlMs: Long = DEFAULT_TTL_MS,
+    private val capacity: Int = DEFAULT_CAPACITY,
+) {
+    private data class Entry(val recording: FrameRecording, val startedAtMs: Long)
+
+    private val lock = Any()
+    private val entries = LinkedHashMap<String, Entry>()
+
+    /** Null without a session: a retry cannot be tied to its first attempt. */
+    fun key(sessionId: String?, upstreamBody: String): String? {
+        val session = sessionId?.takeIf { it.isNotBlank() } ?: return null
+        return "$session:${sha256Hex(upstreamBody)}"
+    }
+
+    /** A compaction's recording, from its first frame: a retry may attach while it is in flight. */
+    fun begin(key: String, recording: FrameRecording) {
+        synchronized(lock) {
+            sweep()
+            entries[key] = Entry(recording, clock())
+        }
+    }
+
+    /** The drive ended. [keep] is the caller's verdict — the client was gone AND the terminal
+     *  ended cleanly (TurnTerminal.endedCleanly); anything else would replay a truncated or failed
+     *  stream to the retry, so it is dropped. A recording a newer begin() replaced is left alone. */
+    fun finish(key: String, recording: FrameRecording, keep: Boolean) {
+        synchronized(lock) {
+            val superseded = entries[key]?.recording !== recording
+            if (!superseded && !keep) entries.remove(key)
+        }
+    }
+
+    fun lookup(key: String): FrameRecording? = synchronized(lock) {
+        sweep()
+        entries[key]?.recording
+    }
+
+    /** A delivered replay has served its purpose; a second identical request runs upstream. */
+    fun consumed(key: String) {
+        synchronized(lock) { entries.remove(key) }
+    }
+
+    private fun sweep() {
+        val now = clock()
+        entries.entries.removeIf { now - it.value.startedAtMs > ttlMs }
+        while (entries.size > capacity) entries.remove(entries.keys.first())
+    }
+
+    private fun sha256Hex(text: String): String =
+        MessageDigest.getInstance("SHA-256")
+            .digest(text.toByteArray(Charsets.UTF_8))
+            .joinToString("") { "%02x".format(it) }
+}
+
+private const val DEFAULT_TTL_MS = 2 * 60 * 60 * 1000L
+private const val DEFAULT_CAPACITY = 32
--- a/gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
+++ b/gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
@@ -61,10 +61,17 @@ internal class CompactionReplay(
         synchronized(lock) { entries.remove(key) }
     }
 
+    // Past capacity, a settled recording goes before one still in flight: begin() inserts at the
+    // START of a compaction, so insertion order alone would evict the oldest-begun entry while its
+    // drive is still running and its retry has not arrived yet (review of PR 137). Only when every
+    // entry is in flight does the oldest go.
     private fun sweep() {
         val now = clock()
         entries.entries.removeIf { now - it.value.startedAtMs > ttlMs }
-        while (entries.size > capacity) entries.remove(entries.keys.first())
+        while (entries.size > capacity) {
+            val victim = entries.entries.firstOrNull { it.value.recording.isComplete }?.key ?: entries.keys.first()
+            entries.remove(victim)
+        }
     }
 
     private fun sha256Hex(text: String): String =
--- a/gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
+++ b/gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
@@ -29,34 +29,54 @@ internal class HeadAdmission(
         val slot = admission.acquireSlotOrRespond(call) ?: return
         telemetry.markAdmitted(perf)
         // A detached compaction takes the slot with it (TurnStreamer.driveDetachable): the drive
-        // releases it when the upstream turn ends, not this call when its client has gone.
-        var handedOff: AtomicBoolean? = null
+        // releases it when the upstream turn ends, not this call when its client has gone. The flag
+        // is this call's from the start, never a return value: the drive's cancelled-call path
+        // THROWS out of serve (review of PR 137), and a finally must not read the flag off a call
+        // that never returned.
+        val admitted = Admitted(slot, t0, perf)
 
         try {
             val prepared = admission.materializeOrRespond(call) { preparation.prepareTurn(call, perf) } ?: return
-            handedOff = serve(call, prepared, slot, t0, perf)
+            serve(call, prepared, admitted)
         } finally {
-            withContext(NonCancellable) { if (handedOff?.get() != true) slot.release() }
+            withContext(NonCancellable) { if (!admitted.handedOff.get()) admitted.slot.release() }
         }
     }
 
-    /** Returns the drive's slot-handoff flag for a driven turn; null for the answers that need none. */
-    private suspend fun serve(
-        call: ApplicationCall,
-        prepared: Preparation,
-        slot: InflightGate.Slot,
-        t0: Long,
-        perf: TurnPerf,
-    ): AtomicBoolean? = when (prepared) {
-        is Preparation.Rejected -> null.also { responses.respondInvalidRequest(call, prepared.message) }
-        is Preparation.Local -> null.also { driver.answerLocally(call, prepared) }
-        is Preparation.Replay -> null.also { driver.replay(call, prepared) }
-        is Preparation.Ready -> {
-            val inputs = TurnInputs(prepared.built, slot, t0, perf)
-            // stream:true → SSE (the interactive path); stream:false → one buffered JSON body
-            // (Claude Code's internal non-stream calls, served by collecting the same machinery).
-            if (prepared.stream) driver.stream(call, inputs) else driver.collect(call, inputs)
-            inputs.slotHandedOff
+    /** What one admitted call holds: its gate slot, its start, its perf row, and the flag a
+     *  detached drive flips when it takes the slot with it. */
+    private data class Admitted(
+        val slot: InflightGate.Slot,
+        val t0: Long,
+        val perf: TurnPerf,
+        val handedOff: AtomicBoolean = AtomicBoolean(false),
+    )
+
+    private suspend fun serve(call: ApplicationCall, prepared: Preparation, admitted: Admitted) {
+        when (prepared) {
+            is Preparation.Rejected -> responses.respondInvalidRequest(call, prepared.message)
+            is Preparation.Local -> driver.answerLocally(call, prepared)
+            is Preparation.Replay -> {
+                // A retry following a compaction still in flight is not an admission: the drive it
+                // follows holds a slot already (TurnStreamer.driveDetachable), so this one goes
+                // back before the wait — else one compaction counts twice against the gate for
+                // however long the upstream turn still runs (review of PR 137). Release is
+                // idempotent (InflightGate.Slot), so the finally above stays a no-op for it.
+                if (!prepared.recording.isComplete) admitted.slot.release()
+                driver.replay(call, prepared)
+            }
+            is Preparation.Ready -> {
+                val inputs = TurnInputs(
+                    prepared.built,
+                    admitted.slot,
+                    admitted.t0,
+                    admitted.perf,
+                    slotHandedOff = admitted.handedOff,
+                )
+                // stream:true → SSE (the interactive path); stream:false → one buffered JSON body
+                // (Claude Code's internal non-stream calls, served by collecting the same machinery).
+                if (prepared.stream) driver.stream(call, inputs) else driver.collect(call, inputs)
+            }
         }
     }
 }
--- a/gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
+++ b/gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
@@ -9,6 +9,9 @@ package splice.gateway.head
 import io.ktor.server.application.ApplicationCall
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.withContext
+import splice.core.perf.TurnPerf
+import splice.spi.InflightGate
+import java.util.concurrent.atomic.AtomicBoolean
 
 internal class HeadAdmission(
     private val deps: HeadDeps,
@@ -25,23 +28,35 @@ internal class HeadAdmission(
         val t0 = deps.clock()
         val slot = admission.acquireSlotOrRespond(call) ?: return
         telemetry.markAdmitted(perf)
+        // A detached compaction takes the slot with it (TurnStreamer.driveDetachable): the drive
+        // releases it when the upstream turn ends, not this call when its client has gone.
+        var handedOff: AtomicBoolean? = null
 
         try {
             val prepared = admission.materializeOrRespond(call) { preparation.prepareTurn(call, perf) } ?: return
-            when (prepared) {
-                is Preparation.Rejected -> responses.respondInvalidRequest(call, prepared.message)
-                is Preparation.Ready -> {
-                    // stream:true → SSE (the interactive path); stream:false → one buffered JSON body
-                    // (Claude Code's internal non-stream calls, served by collecting the same machinery).
-                    if (prepared.stream) {
-                        driver.stream(call, prepared.built, slot, t0, perf)
-                    } else {
-                        driver.collect(call, prepared.built, slot, t0, perf)
-                    }
-                }
-            }
+            handedOff = serve(call, prepared, slot, t0, perf)
         } finally {
-            withContext(NonCancellable) { slot.release() }
+            withContext(NonCancellable) { if (handedOff?.get() != true) slot.release() }
+        }
+    }
+
+    /** Returns the drive's slot-handoff flag for a driven turn; null for the answers that need none. */
+    private suspend fun serve(
+        call: ApplicationCall,
+        prepared: Preparation,
+        slot: InflightGate.Slot,
+        t0: Long,
+        perf: TurnPerf,
+    ): AtomicBoolean? = when (prepared) {
+        is Preparation.Rejected -> null.also { responses.respondInvalidRequest(call, prepared.message) }
+        is Preparation.Local -> null.also { driver.answerLocally(call, prepared) }
+        is Preparation.Replay -> null.also { driver.replay(call, prepared) }
+        is Preparation.Ready -> {
+            val inputs = TurnInputs(prepared.built, slot, t0, perf)
+            // stream:true → SSE (the interactive path); stream:false → one buffered JSON body
+            // (Claude Code's internal non-stream calls, served by collecting the same machinery).
+            if (prepared.stream) driver.stream(call, inputs) else driver.collect(call, inputs)
+            inputs.slotHandedOff
         }
     }
 }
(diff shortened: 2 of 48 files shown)


gateway/gateway/src/main/kotlin/splice/gateway/head/TurnPreparation.kt
```diff
@@ -61,6 +98,37 @@ internal class TurnPreparation(
...
         perf.mark(PerfKeys.BUILD)
-        return Preparation.Ready(built, parsed.typed.stream)
+        // A compaction retry whose bytes match a compaction that outlived its first client is
+        // answered from that recording (TurnStreamer records it, LocalResponses replays it).
+        val replayed = if (built.meta.compact && parsed.typed.stream) replayFor(built) else null
+        return replayed ?: Preparation.Ready(built, parsed.typed.stream)
     }
+
+    private fun replayFor(built: BuiltTurn): Preparation.Replay? {
+        val key = replay.key(built.meta.sessionId, built.requestBody.toString()) ?: return null
```
marcospaulo, line 127: **[concern]** The `keep` verdict in `CompactionReplay.finish()` is the right instinct, and `if (!superseded && !keep) entries.remove(key)` (`CompactionReplay.kt:50`) does close the case it was written for: a failed or truncated recording never answers a *future* lookup. There is one retry it cannot reach, though, and it is the one that arrives at the worst moment.

`replayFor()` attaches a retry to a recording that is still being written:

```kotlin
val recording = replay.lookup(key) ?: return null
val state = if (recording.isComplete) "finished" else "still running"
```
(`TurnPreparation.kt:108-110`)

Following that chain in this PR: `LocalResponses.replay()` calls `replayed.recording.follow { ... }` (`LocalResponses.kt:63-72`), and `follow()` parks on `progress.first { it.frames > seen || it.complete }` (`FrameRecording.kt:48`) until the live turn calls `complete()`. Meanwhile `ClientChannel.timedClientWrite` appends every frame to the recording *before* it checks detachment:

```kotlin
fun timedClientWrite(frame: String, perf: TurnPerf, clock: ElapsedClock) {
    recording?.append(frame)
    if (detached.get()) return
```
(`ClientChannel.kt:68-70`)

So the cancellation seal's honest `event: error` frame lands in the same recording, and the already-attached retry is streamed it verbatim. `finish(keep = false)` removes the map entry, which does nothing for a follower already inside `follow()`.

Falsifiable claim: if a detached compaction ends with `endedCleanly == false` (head stop, upstream error) while a retry is parked in `follow()`, that retry receives the recorded error frame instead of the fresh upstream attempt it would have had before this PR. A direct test: `begin()` a recording, start a follower, append an `event: error` frame, `complete()`, `finish(keep = false)`, and assert what the follower received.

The window is not far from the motivating incident: the retry is sent near the client's own 600 s recovery timeout, which is roughly when the original attempt is most likely to be failing too.

One direction, if the simple shape is acceptable: only attach a retry to a recording whose outcome is already known good.

```kotlin
private fun replayFor(built: BuiltTurn): Preparation.Replay? {
    val key = replay.key(built.meta.sessionId, built.requestBody.toString()) ?: return null
    // A live turn can still fail, and a follower already inside follow() cannot be recalled,
    // so only a settled recording answers a retry; anything else runs upstream as before.
    val recording = replay.lookup(key)?.takeIf { it.isComplete } ?: return null
    ...
}
```

That trades away the "retry rides the live turn" behavior, which may be deliberate. If you want to keep it, the other direction is to have `follow()` surface the terminal verdict so the replay can abandon mid-stream and fall back to `Preparation.Ready`. I might be off on which of the two you prefer here, but the current unconditional attach is the one shape that cannot recover.

This also closes the double slot hold flagged in the separate comment on `HeadAdmission.kt:53`: same trigger (a retry attaching to an in-flight recording), different failure.

gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
```diff
@@ -25,23 +28,35 @@ internal class HeadAdmission(
...
+    private suspend fun serve(
+        call: ApplicationCall,
+        prepared: Preparation,
+        slot: InflightGate.Slot,
+        t0: Long,
+        perf: TurnPerf,
+    ): AtomicBoolean? = when (prepared) {
+        is Preparation.Rejected -> null.also { responses.respondInvalidRequest(call, prepared.message) }
+        is Preparation.Local -> null.also { driver.answerLocally(call, prepared) }
+        is Preparation.Replay -> null.also { driver.replay(call, prepared) }
```
marcospaulo: **[concern]** Routing `Preparation.Replay` through the same admission path as everything else keeps this branch simple, and `LocalResponses`' header states the intended contract plainly: "Both hold the admission slot for their own duration only." That contract holds for a finished recording. It does not hold for a live one.

```kotlin
is Preparation.Replay -> null.also { driver.replay(call, prepared) }
```
(`HeadAdmission.kt:53`)

`also` is inline and this branch runs inside `handleMessages`' `try`, so the retry's own slot (acquired at `HeadAdmission.kt:29`) is held for the whole of `driver.replay`. When the matched recording is still in flight, `follow()` does not return until the original detached drive calls `recording.complete()` (`TurnStreamer.kt:131`), and that drive releases its own slot one line later:

```kotlin
replay.finish(key, recording, keep = kept)
inputs.slot.release()
```
(`TurnStreamer.kt:137-138`)

Falsifiable claim: for the whole overlap, one logical compaction holds two `InflightGate` slots, and the duration of that overlap is however long the live upstream turn still has to run (minutes, in the incident this PR is built around). Against a `MAX_INFLIGHT` default of 12 (`Knob.kt`), a handful of concurrent retry-versus-live pairs measurably shrinks real headroom, exactly under the load condition the PR is trying to relieve. It is also invisible in `AdmissionTelemetry`, since both holds look like ordinary admissions.

Two directions, depending on which invariant you would rather keep:

```kotlin
// (a) the replay of a LIVE recording is a follower, not an admission: give the slot back
//     before parking on follow(), since the drive it is following already holds one.
is Preparation.Replay -> null.also {
    if (!prepared.recording.isComplete) slot.release()
    driver.replay(call, prepared)
}
```

or (b) refuse to attach to an in-flight recording at all, which is the fix suggested in the comment on `TurnPreparation.kt:108` and removes this window as a side effect.

If neither is worth the change, pricing it explicitly (a counter for "replays following a live drive") would at least keep the double accounting visible rather than silent.

gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /** A delivered replay has served its purpose; a second identical request runs upstream. */
+    fun consumed(key: String) {
+        synchronized(lock) { entries.remove(key) }
+    }
+
+    private fun sweep() {
+        val now = clock()
+        entries.entries.removeIf { now - it.value.startedAtMs > ttlMs }
+        while (entries.size > capacity) entries.remove(entries.keys.first())
```
marcospaulo: **[suggestion]** The TTL sweep and the capacity bound are both worth having, and pairing them in one `sweep()` keeps the map honest without a background task. One detail in the capacity half is worth a guard.

```kotlin
while (entries.size > capacity) entries.remove(entries.keys.first())
```
(`CompactionReplay.kt:67`)

`entries` is a plain `LinkedHashMap` (`CompactionReplay.kt:28`), so it is insertion-ordered and re-inserting an existing key does not move it. `begin()` inserts at the *start* of a detached compaction (`CompactionReplay.kt:37-42`), which means "oldest" here is oldest-begun, not oldest-finished. A compaction that is still driving (`recording.isComplete == false`) is evicted on exactly the same terms as one that settled an hour ago.

Falsifiable claim: with more than 32 concurrently detached compactions on one head, the oldest still-running one is dropped from the map, its job keeps running (it does not know it was evicted), and a retry that arrives afterwards misses in `lookup()`, falls through to `Preparation.Ready`, and pays for a second upstream compaction for the same session. `DEFAULT_CAPACITY` is 32 (`CompactionReplay.kt:77`) while `Knob.kt` documents raising `maxInflight` above the default 12 for high-capacity backends, so this is reachable by configuration rather than only in theory.

Cheapest fix that keeps the bound meaningful is to prefer settled entries when evicting:

```kotlin
private fun sweep() {
    val now = clock()
    entries.entries.removeIf { now - it.value.startedAtMs > ttlMs }
    while (entries.size > capacity) {
        // A still-driving compaction has a retry coming; drop a settled one first, and only
        // fall back to the oldest when everything held is still in flight.
        val victim = entries.entries.firstOrNull { it.value.recording.isComplete }?.key
            ?: entries.keys.first()
        entries.remove(victim)
    }
}
```

Sizing `capacity` off `maxInflight` instead of a literal 32 would work too, and is arguably the more honest coupling since both bound the same thing.

gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
```diff
@@ -25,23 +28,35 @@ internal class HeadAdmission(
...
-                    // stream:true → SSE (the interactive path); stream:false → one buffered JSON body
-                    // (Claude Code's internal non-stream calls, served by collecting the same machinery).
-                    if (prepared.stream) {
-                        driver.stream(call, prepared.built, slot, t0, perf)
-                    } else {
-                        driver.collect(call, prepared.built, slot, t0, perf)
-                    }
-                }
-            }
+            handedOff = serve(call, prepared, slot, t0, perf)
```
marcospaulo: **[concern]** The hand-off design is sound and the comment at `HeadAdmission.kt:31-32` says exactly what it does. This is about how the flag travels, not about whether it works today, so leading with what I checked: I could not make this fail. Running the PR-head test `a compaction whose client hangs up finishes detached and its answer is replayed to the retry` with a `delay(1_000)` inserted before the inflight assertion still passes (1 test, 3.376 s, 0 failures), so a full second after the "call cancelled" log line the gate still holds the drive's slot. And `InflightGate.Slot.release()` is guarded by `released.compareAndSet(false, true)`, so a second release can never decrement the gate twice. There is no double release here.

What is left is a fragility worth a comment. The cleanup decision travels as the return value of a call that can throw:

```kotlin
handedOff = serve(call, prepared, slot, t0, perf)
} finally {
    withContext(NonCancellable) { if (handedOff?.get() != true) slot.release() }
}
```
(`HeadAdmission.kt:37-39`)

The flag is produced only as the last expression of `serve()`'s `Ready` branch (`inputs.slotHandedOff`, `HeadAdmission.kt:59`), after the `driver.stream(call, inputs)` call on line 58. `TurnStreamer.driveDetachable` sets `inputs.slotHandedOff.set(true)` at line 123 and rethrows on the cancelled-call path at line 152. So the flag is already true at a moment when `handedOff` is still null, and the finally would branch on `null != true`.

Falsifiable claim, stated as the unobserved thing it is: if any throw ever escapes `respondTextWriter` after line 123 has run, the slot returns to the gate early (once, not twice) while the detached drive is still running upstream, and the drive's own `release()` at `TurnStreamer.kt:138` becomes the no-op. The gate would then be one admission light for the rest of that turn. I have not reproduced it, and the `throw e` path at `TurnStreamer.kt:152` does not appear to be the one `socket.close()` exercises in the test.

The cheap way to make it structurally impossible is to hand the flag out before the call that can throw, rather than after it:

```kotlin
private suspend fun serve(
    call: ApplicationCall,
    prepared: Preparation,
    slot: InflightGate.Slot,
    t0: Long,
    perf: TurnPerf,
    onHandedOff: (AtomicBoolean) -> Unit,
) = when (prepared) {
    ...
    is Preparation.Ready -> {
        val inputs = TurnInputs(prepared.built, slot, t0, perf)
        onHandedOff(inputs.slotHandedOff)   // before anything that can throw
        if (prepared.stream) driver.stream(call, inputs) else driver.collect(call, inputs)
    }
}
```

Separately, and this one is testable now: the test's `logged("compaction continues detached")` cannot tell the cancellation path from the keepalive path, because `DETACHED_NOTE` at `ClientChannel.kt:49` is the same string `TurnStreamer.kt:148` logs. A green run of that test is not evidence that the `job.join()` cancellation branch is covered. Giving the two sites distinguishable text (or asserting on the `call cancelled` fragment specifically) would make the assertion mean what it looks like it means.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnPreparation.kt
```diff
@@ -61,6 +98,37 @@ internal class TurnPreparation(
             prepared
         }
         perf.mark(PerfKeys.BUILD)
-        return Preparation.Ready(built, parsed.typed.stream)
+        // A compaction retry whose bytes match a compaction that outlived its first client is
+        // answered from that recording (TurnStreamer records it, LocalResponses replays it).
+        val replayed = if (built.meta.compact && parsed.typed.stream) replayFor(built) else null
```
marcospaulo: **[suggestion]** Gating the new machinery on `compact` is right, and keeping ordinary turns on the untouched path is what makes this PR safe to land. The second half of the condition is the part I would like to see stated somewhere in the code.

```kotlin
val replayed = if (built.meta.compact && parsed.typed.stream) replayFor(built) else null
```
(`TurnPreparation.kt:103`)

The `stream` half also gates the detached drive, not just the replay lookup: `driveDetachable` is only reachable from `TurnStreamer.stream()`, and `CollectTurn.kt` has no match for `recording`, `detach`, `replay` or `slotHandedOff`. So a compact turn with `stream:false` is routed by `HeadAdmission.kt:58` to `driver.collect` and gets none of this: no detached scope, no recording, no replay.

Falsifiable claim: if a client ever issues a compaction with `stream:false`, a client-side abandonment of it cancels the call exactly as it did before 2026-09-05, and nothing in the logs says the protection did not apply. Today Claude Code's auto-compaction is `stream:true`, so this is a boundary rather than a live defect, which is why this is a suggestion and not a concern. What makes it worth a line of code is that the boundary is currently silent.

The smallest useful version is to say so where the decision is made, and to notice if the assumption ever breaks:

```kotlin
// The detached drive and the replay lookup are both stream-only: driveDetachable is reachable
// only from TurnStreamer.stream(), and CollectTurn has no recording path. A non-stream
// compaction therefore behaves as every turn did before 2026-09-05.
val replayed = if (built.meta.compact && parsed.typed.stream) {
    replayFor(built)
} else {
    if (built.meta.compact) deps.log("[${provider.key}] non-stream compaction: no detached drive\n")
    null
}
```

If extending the protection to `collect` is on the roadmap anyway, a `TODO` naming that is fine too. The point is that the next person reading `TurnPreparation` should not have to trace `CollectTurn` to learn the feature is half-covered by design.

gateway/gateway/src/test/kotlin/head/HeadServerLocalAnswerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+            setBody(body(stream))
+        }.bodyAsText()
+
+    @Test
+    fun `the streamed side query gets a one-block end_turn answer and upstream sees nothing`() = runBlocking {
+        val before = mock.upstreamBodies.size
+        val sse = post(stream = true)
+        assertTrue(sse.contains("event: message_start"), sse)
+        assertTrue(sse.contains("Reading runAgent.ts"), sse)
+        assertTrue(sse.contains("\"stop_reason\":\"end_turn\""), sse)
```
marcospaulo, line 130: **[suggestion]** These two tests pin the important behavioral claim well: the label text is right and `mock.upstreamBodies.size` is unchanged, which is the assertion that actually proves the side query never reaches upstream. The gap is on the other risk the PR names for this same code.

The PR's own risk register says the synthesized response "must be wire-exact (usage, stop_reason, ids) or the client HUD and accounting drift", and `LocalResponses.answer` does synthesize a usage payload (`wiring.usagePayloadBuilderFor`, `LocalResponses.kt:31-35`). But the assertions stop at the text and the stop reason:

```kotlin
assertTrue(sse.contains("\"stop_reason\":\"end_turn\""), sse)
```
(`HeadServerLocalAnswerTest.kt:130`, and the same shape at line 144)

Neither the streamed test (lines 124-137) nor the non-stream one (lines 139-146) reads the usage block or any id field.

Falsifiable claim: a regression that emitted the locally composed response with a missing `usage` object, or with an id that does not match the shape Claude Code expects, passes both tests untouched, even though that is the failure mode the PR itself flags as the one that drifts the HUD and accounting.

Two assertions per test would cover it:

```kotlin
assertTrue(sse.contains("\"usage\":"), "the synthesized answer must carry a usage block: $sse")
assertTrue(
    Regex("\"id\":\"msg_[A-Za-z0-9_-]+\"").containsMatchIn(sse),
    "the message id must match the shape the client parses: $sse",
)
```

Pinning the actual token numbers is probably over-fitting given the payload builder owns them, but pinning that the block exists and that the id has the right prefix would catch the drift class without freezing the values.

marcospaulo (commented):
**Review of #137** (head 33bde532): a compaction turn survives its client and is replayed to the byte-identical retry, the activity-label side query is answered locally, codex-rs routing headers ride every turn, learned client windows persist, and every OAuth head owns a splice credential file.

**Findings: 6 inline comments** (2 concern, 4 suggestion). Raw findings: 48 (40 mechanical, 8 from five lenses and a flow map); 42 dissolved with a named gate (34 dormant-pack rule hits judged false positives on their instance, 8 template comments).

**Where the concerns cluster:** the retry-attach seam of the new replay path. A retry that attaches to a compaction still in flight follows the live recording and is handed that compaction's eventual error frame if it fails (TurnPreparation), and holds its own admission slot for the whole follow, so one logical compaction counts twice against the head's inflight limit (HeadAdmission). Suggestions: capacity eviction can drop an in-flight entry (CompactionReplay), a non-stream compact turn gets none of the new protection (TurnPreparation), the admission finally reads the handoff flag off a call's return value rather than the inputs it already holds (HeadAdmission, fragility only: a probe run with a 1 s delay before the inflight assertion showed the slot stays with the drive after a hangup, and slot release is compare-and-set guarded), and the local-answer tests do not pin usage or ids (HeadServerLocalAnswerTest).

**Checked and found clean:** every OAuth default-file resolution site reads the registry (login, doctor, chat arm, kimi, both knobs); doctor warns only on an explicit vendor-file path so upgraders get the missing-credential path, not a silent sign-out; ClientWindows persists via temp file plus atomic move and rejects non-positive values on load; no request text or full session id reaches the daemon log on the new paths; ActivityLabel matches only the last user message's verbatim opening sentence, with three near-miss inputs pinned; codex routing headers are a pure function of session and model, constant across a compaction; the detached scope survives a head stop and start, and the ATOMIC launch guarantees the drive's finally; no CI workflow or dependency file is touched; the example config's three OAuth blocks are consistent with the README.

**Coverage disclosure:** three files were covered by the mechanical scan only (CHANGELOG.md, README.md, StatusCommandTest.kt). Mechanical scan: run, 94 rules, 2 skipped (sgconfig.yml, detekt.yml); all 40 hits came from the repo's dormant pack, none from the enforced walls. Test-lens notes with no anchorable line: no end-to-end test drives the mock upstream to failure while a compaction is detached, and no HeadServer-level test posts the retry while the live turn is still running.

**Optional follow-up (not installed):** an ast-grep antibody sketch `kt-finally-flag-from-throwing-call` for a cleanup flag read in `finally` from a var assigned only by a throwing call inside `try`; it needs tightening before it could run in the enforced pack.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnPreparation.kt
marcospaulo: Fixed in 46afa903, keeping the attach-to-live design (in the motivating case the retry arrives while the 5 to 10 minute compaction is still running, and riding it is the whole saving). The gap was narrower than the comment's first paragraph: an upstream failure mid-turn already records an error frame the follower forwards, so the one path that left a follower with a truncated 200 was a cancelled detached drive, where the seal abandons because the drive's own client is gone. Now the drive's terminal verdict travels with the recording (`FrameRecording.complete(whole)`, and `follow()` returns it), and `LocalResponses.replay` seals a follower whose recording ends not whole with the same overloaded error frame the seal writes for an attached client. That client retries, and finds the entry gone (`finish(keep = false)`), so the next attempt runs upstream. Pinned at the recording level (`FrameRecordingTest`: a follower gets the drive's verdict back). One honest limit: with today's stop ordering (drain, engine stop, then cancel detached drives) the follower's connection is torn down before the seal could reach it, so the seal is the correct ending for any future cancellation path rather than an observed repair.

gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
marcospaulo: Fixed in 46afa903, direction (a): a retry that attaches to a recording still in flight releases its own slot before parking on `follow()`; the drive it follows holds one already, and `Slot.release()` is idempotent so the call's finally stays a no-op for it. Pinned by a new HeadServer test (`a retry that arrives while the compaction is still in flight follows it on the drive's slot`): with the upstream parked and the retry following, `inflight` is 1, and the retry receives the whole answer from the one upstream turn. Red on the old code with `inflight=2`.

gateway/gateway/src/main/kotlin/splice/gateway/head/CompactionReplay.kt
marcospaulo: Fixed in 46afa903 as suggested: past capacity, `sweep()` evicts the first settled entry and falls back to the oldest only when every entry is still in flight. Pinned by `CompactionReplayTest` (`past capacity a settled recording goes before one still in flight`), red on the old code (the live entry was gone).

gateway/gateway/src/main/kotlin/splice/gateway/head/HeadAdmission.kt
marcospaulo: Fixed structurally in 46afa903: the handoff flag is the admitted call's from the start (a small `Admitted` holder of slot, start, perf row and flag) and is threaded into `TurnInputs`, so `serve()` returns nothing and the finally never reads a value off a call that may have thrown. On the test assertion: it stays on the shared `DETACHED_NOTE` on purpose. On a real FIN, Ktor's cancellation and the keepalive pinger's failed write race, both detach, and this run saw each path win in different tests (the first cut of the new test asserted "call cancelled" and failed because the pinger got there first). The fact under test is that the compaction detached and kept its slot, which holds for both paths.

gateway/gateway/src/main/kotlin/splice/gateway/head/TurnPreparation.kt
marcospaulo: Done in 46afa903: the decision moved into `compactionReplay(built, stream)`, whose doc states that the detached drive and the replay are stream-only (the drive lives in `TurnStreamer.stream()`, `CollectTurn` has no recording), and a compaction with `stream:false` logs `non-stream compaction (... served attached: no detached drive, no replay)` so the boundary is visible the day a client crosses it.

gateway/gateway/src/test/kotlin/head/HeadServerLocalAnswerTest.kt
marcospaulo: Added in 46afa903 to both tests: the response carries a `usage` block, and the message id matches the shape `MessageIds` mints (`msg_<millis>_<seq>`), pinned as presence and shape rather than values, as suggested.

PR #153, opened 2026-09-15
fix(login): the device-login test no longer opens browser tabs

marcospaulo:
## Summary
- `DeviceLoginTokenlessTest` fed the login flow a fake device response and the flow opened its verification URL in the operator's real browser: four `xdg-open` tabs on `http://127.0.0.1/verify` on every test run.
- The browser opener is now a `fun interface` seam on `LoginIo` with the operating-system process as the default; `DeviceLoginFlow` threads a per-call `LoginIo` through an internal `run` overload (public API unchanged); the test injects a recording opener and pins the exact URL and the manual fallback line.
- Same change already lands on `feat/v0.4.0`; this carries it to `main` so every checkout stops opening tabs.

## Test plan
- [x] `:app:test` with an `xdg-open` shim on PATH: 4 launches on `main` before the change, 0 after; 340 app tests, 0 failed
- [x] `:app:detekt`, `gate:rules` 47/47, `git diff --check` clean
- [ ] CI on the PR merge ref


marcospaulo:
Superseded by the v0.4.0 PR: this change is already contained in feat/v0.4.0 (DeviceLoginTokenlessTest drives DeviceLoginFlow with an injected LoginIo(browser) there too). Closing so there is one branch and one PR.

PR #154, opened 2026-09-17
feat: splice v0.4.0 — everything but the console

marcospaulo:
## splice v0.4.0 — everything but the console

The whole `feat/v0.4.0` campaign (ledger: `dev/campaigns/v0.4.0.toml`, every row `verified` by the orchestrator with its gate of record) consolidated into one branch and one PR.

### What ships
- **Setup, drift warning, doctor --json, `splice add`, `splice upgrade`**, perf/sessions data, compaction instructions, shared MCP hosting, local models, account pools with automatic switching, code mode on by default for Claudex-shaped providers, the Muse head.
- **Retry, end to end (V4-50..V4-74):** every upstream status backs off on one curve; a persistent 429 is refused at admission with a real 429, `Retry-After` and the `anthropic-ratelimit-unified-*` headers; the first 429 turn, an empty model turn and an empty compaction all reach the client as the one in-band error it retries; every launched client runs in persistent retry mode and sleeps until the reset splice sends; a spent account (grok spending-limit 403, any 402) is a rate limit to every layer; pre-content stream tears are continued; a daemon restart drains in-flight turns instead of cutting them (V4-74); a stream cut after a completed tool call ends the turn clean at `tool_use` instead of an error the client can only finalize as "Connection lost" (V4-76).
- **Every in-band error is retryable before content (V4-78, V4-79):** Claude Code retries an in-band error event only when its body says `overloaded_error`, so a pre-content `api_error` was terminal. One rule (`PreContentWireType`) now wires every pre-content `api_error`/`rate_limit_error` as `overloaded_error` at every emitter (known-end, the internal-error site, the oversized-event site, the pipeline finish), and a campaign wall fails by name on any emitter that bypasses it.
- **Pooled-account exhaustion is a bounded refusal (V4-77, V4-80):** `Retry-After` is the cooldown lift (≤120s), never the provider's reset window (up to 7 days), and the refusal carries the same `rejected` unified headers as a cooldown refusal.
- **Adversarial re-review of every landed row:** 25 rows marked done by builders were independently reviewed; five failed and were redone — session cost double-billed cached tokens (46.47 USD shown for a 1.42 USD session; V4-37), a synthesized grok expiry could veto a genuine refresh for 4h (V4-38), `splice add-model` never reached `/v1/models` on the shipped starter (V4-34), the per-head system prompt shipped without its example/README/doctor WARN (V4-36), and the heads e2e planted a fake MCP server (now a real stdio server, enabled in the scratch settings; V4-33).
- **Post-review hardening (V4-81..V4-86, from `/code-review` on the branch):** the pre-content wire-type rule now lives once at the emitter seam, knows permanence (a deterministic refusal or an unparseable base_url keeps its real type instead of becoming a 300-retry overloaded_error storm), and leaves the non-streaming path on its real statuses; the wall enforces that one seam. A pooled refusal ships the selected account's quota headers. `splice add-model` edits the roster through the TOML structure mask, re-parses before writing, and refuses as a sentence. Cache-write tokens are their own bucket from the passthrough usage through session cost, doctor, the hourly economics rollup and the burn page.
- **Auth hardening:** a re-login can no longer be mistaken for the credential it replaced; a billing rejection never reopens the login page.
- **README** rewritten for the wider provider set; the kimi failure golden now freezes the wire, not an internal rendering.

### Consolidation
- Supersedes #153 (`fix/device-login-browser-seam`): its change is already contained here byte-for-byte (`DeviceLoginTokenlessTest` uses the injected `LoginIo(browser)`).
- `feat/burn-dashboard` (the economics tab) is folded in as a REBUILD (V4-75): it forked 49 commits ago and a cherry-pick conflicted structurally, so the hourly rollup, `/api/economics` and the burn tab were re-derived against today's seams and the branch is deleted.

### Architecture audit → walls → fixes (in progress on this branch)
A four-lane architecture audit (layering, unidirectional flow and seams, TOML configurability, risk) was run on 2026-09-17. Per concept #924 every violation class became an instrument BEFORE its instances were fixed: 14 new ast-grep walls (routed into the PreToolUse hook and `gate:rules` from one source), 9 python gate legs with red-green selftests (knob/env documentation denominators parsed from source, constant single-source with a strict scar plane and a COPY/COLLISION ratchet, silent-constants ratchet, role registry, model catalogs, autocloseable-closed, public-surface and constructor-width ratchets, schema-keys-consumed), Konsist laws (core OS-escape denylist, Konsist map read from the Gradle module law, stale-allowance detection) and two behaviour walls (DaemonStopOrderTest, economics contract derived from the row type). Fix rows V4-99..V4-115 burn the inventories down; the gate is red BY NAME until they land.

### Head isolation (V4-115, operator ruling 2026-09-17)
The 2026-09-16 cross-head resume linked every head's `projects/` at `~/.claude/projects`, so vanilla `claude -c` could restore a claude-deepseek transcript and warn about an unknown model. Sessions become bounded by head (real per-head trees, no shared link); cross-head resume is an explicit `-r <session-id>` copy into the calling head's tree with the model fields rewritten to that head's pinned model.

### The last three feature rows (V4-175 … V4-177)

- **V4-175 — the wizard asks which lane, and the console reads the payload the daemon sends.** `webui`'s claude-head slice had been written against a payload `/api/claude-head` was expected to serve and never did: `head`, `config_dir`, `auth_kind`, `claude_on_path`, `wrap_supported` were declared and never crossed the wire, while `resolves_to`, `real_binary_path` and `claude_logins` crossed it and nothing read them. It survived because every check took its denominator from the declaration it was checking, and `/api/claude-head` was the one route missing from the daemon's `WebuiContractTest` — now its eighth arm, so the field names are pinned from outside. Two shared bugs fell out: the console lifted `error.message` from a control plane that writes `{"error": "<sentence>"}`, so every daemon refusal reached the screen as `HTTP 409`; and a wrap action returned its own answer instead of re-reading, so a refused wrap left the page claiming a lane that was not in force. `splice setup` now asks the lane question and drives the same `POST /api/claude-head/wrap`.
- **V4-176 — the supervision contract is stated as config, not as one box's tool name.** 25 references to this machine's private supervisor and build wrapper were spread across 12 files, each written while the fact was fresh and true here. A sweep fixes the 25; the 26th arrives in the next comment. So the row restates the requirement as two knobs (`supervisorUnit`, `mcpSlice` — a box whose packager named the unit differently was reading as permanently unsupervised, which is not cosmetic: `POST /api/daemon/restart` refuses to drain an unsupervised daemon) and adds a wall that fails by `file:line`, mutation-proven on the real tree. The wall tripped on its own explanation's first draft, which quoted the name it bans — that is the check covering its own directory.
- **V4-177 — splice's state root is its own, and a pre-0.4 root is adopted in place.** `~/.claude-codex` was never chosen for splice: it is the isolated config dir of the upstream Node proxy this was ported from, kept deliberately because the paths were an external contract for an out-of-repo HUD. That consumer is gone, measured rather than argued — the live statusline names none of those files, this directory, or splice at all. The rename is one line; the work is the migration. When `~/.splice/state` does not exist and `~/.claude-codex/state` does, the old root IS the state dir, read in place: nothing copied, nothing moved, nothing to half-finish, and the decision is re-derived from the filesystem on every construction so the daemon, the CLI and `splice-launch` agree without coordinating. Both arms key off `<root>/state` and never the root, because `~/.claude-codex` is ALSO the wrapped Claude Code config dir of the head keyed `codex` — probing the root would mint the mgmt-key inside another product's config dir. `splice doctor` says which root is live and warns when a pre-0.4 one is left behind unread; `kt-state-paths-single-source` now guards both roots; and `StateDirAgreementTest` enumerates every bash copy of the rule from the tree and drives all four filesystem shapes plus the env precedence through both implementations, with a planted wrong resolver proving the harness runs the script.

### Gate
`bash checks/gate.sh` in `buildgate.slice` on the clean worktree at the head commit (`5b25e335`, 200 legs): `GATE: PASS` (gradle clean check, ast-grep walls, concentration, campaign walls, oracle replay 11/11, code-mode selftests, config guard, safe-failure-render, webui lint/test/build, OSS readiness, pr-title).

### The deferred row, landed (V4-141)

- **V4-141 — TCP keepalive on upstream sockets, via the OkHttp engine.** `java.net.http.HttpClient` exposes no socket options (JDK-8338681), so for as long as it was the upstream engine a half-open upstream connection read as healthy for the whole turn cap — measured, not argued: the daemon's established upstream sockets showed no keepalive timer. The engine is now ktor's OkHttp, for the one reason that it takes a `SocketFactory`: every upstream socket is armed before it connects (SO_KEEPALIVE, TCP_NODELAY, idle/interval/count = 30s/10s/3, a dead peer known within 60s), which also closes G26. The swap's two costs are paid and pinned rather than met in production. OkHttp's Dispatcher caps five requests per host by default, which would serialise every turn against one provider; it runs at 4096 on a virtual-thread executor. And ktor's OkHttp engine reads each response body with a blocking `source.read` on its engine dispatcher, `Dispatchers.IO` by default — the first run of the swap deadlocked the 1000-stream load test at exactly the ramp width (96 held at the mock, 0 live at the client) because 96 held streams pinned all 64 IO threads and the head could not forward the deltas it already had. The engine dispatcher is the same virtual-thread executor, both are pinned by `UpstreamTransportOkHttpTest`, and the load test is green again at the same concurrency (`n=1000 all_live=true peak_concurrent=1000 clean_finishes=1000 errored=0`). Six components entered in `verification-metadata.xml` by hand, every sha256 cross-checked against Maven Central.


marcospaulo:
Closed by GitHub when `main` was force-pushed to the rewritten history; continues as #157 (same branch, new SHAs).

PR #157, opened 2026-09-20
feat: splice v0.4.0 — everything but the console

marcospaulo:
## splice v0.4.0 — everything but the console

The whole `feat/v0.4.0` campaign (ledger: `.dev/campaigns/v0.4.0.toml`, every row `verified` by the orchestrator with its gate of record) consolidated into one branch and one PR.

### What ships
- **Setup, drift warning, doctor --json, `splice add`, `splice upgrade`**, perf/sessions data, compaction instructions, shared MCP hosting, local models, account pools with automatic switching, code mode on by default for Claudex-shaped providers, the Muse head.
- **Retry, end to end (V4-50..V4-74):** every upstream status backs off on one curve; a persistent 429 is refused at admission with a real 429, `Retry-After` and the `anthropic-ratelimit-unified-*` headers; the first 429 turn, an empty model turn and an empty compaction all reach the client as the one in-band error it retries; every launched client runs in persistent retry mode and sleeps until the reset splice sends; a spent account (grok spending-limit 403, any 402) is a rate limit to every layer; pre-content stream tears are continued; a daemon restart drains in-flight turns instead of cutting them (V4-74); a stream cut after a completed tool call ends the turn clean at `tool_use` instead of an error the client can only finalize as "Connection lost" (V4-76).
- **Every in-band error is retryable before content (V4-78, V4-79):** Claude Code retries an in-band error event only when its body says `overloaded_error`, so a pre-content `api_error` was terminal. One rule (`PreContentWireType`) now wires every pre-content `api_error`/`rate_limit_error` as `overloaded_error` at every emitter (known-end, the internal-error site, the oversized-event site, the pipeline finish), and a campaign wall fails by name on any emitter that bypasses it.
- **Pooled-account exhaustion is a bounded refusal (V4-77, V4-80):** `Retry-After` is the cooldown lift (≤120s), never the provider's reset window (up to 7 days), and the refusal carries the same `rejected` unified headers as a cooldown refusal.
- **Adversarial re-review of every landed row:** 25 rows marked done by builders were independently reviewed; five failed and were redone — session cost double-billed cached tokens (46.47 USD shown for a 1.42 USD session; V4-37), a synthesized grok expiry could veto a genuine refresh for 4h (V4-38), `splice add-model` never reached `/v1/models` on the shipped starter (V4-34), the per-head system prompt shipped without its example/README/doctor WARN (V4-36), and the heads e2e planted a fake MCP server (now a real stdio server, enabled in the scratch settings; V4-33).
- **Post-review hardening (V4-81..V4-86, from `/code-review` on the branch):** the pre-content wire-type rule now lives once at the emitter seam, knows permanence (a deterministic refusal or an unparseable base_url keeps its real type instead of becoming a 300-retry overloaded_error storm), and leaves the non-streaming path on its real statuses; the wall enforces that one seam. A pooled refusal ships the selected account's quota headers. `splice add-model` edits the roster through the TOML structure mask, re-parses before writing, and refuses as a sentence. Cache-write tokens are their own bucket from the passthrough usage through session cost, doctor, the hourly economics rollup and the burn page.
- **Auth hardening:** a re-login can no longer be mistaken for the credential it replaced; a billing rejection never reopens the login page.
- **README** rewritten for the wider provider set; the kimi failure golden now freezes the wire, not an internal rendering.

### Consolidation
- Supersedes #153 (`fix/device-login-browser-seam`): its change is already contained here byte-for-byte (`DeviceLoginTokenlessTest` uses the injected `LoginIo(browser)`).
- `feat/burn-dashboard` (the economics tab) is folded in as a REBUILD (V4-75): it forked 49 commits ago and a cherry-pick conflicted structurally, so the hourly rollup, `/api/economics` and the burn tab were re-derived against today's seams and the branch is deleted.

### Architecture audit → walls → fixes (in progress on this branch)
A four-lane architecture audit (layering, unidirectional flow and seams, TOML configurability, risk) was run on 2026-09-17. Per concept #924 every violation class became an instrument BEFORE its instances were fixed: 14 new ast-grep walls (routed into the PreToolUse hook and `gate:rules` from one source), 9 python gate legs with red-green selftests (knob/env documentation denominators parsed from source, constant single-source with a strict scar plane and a COPY/COLLISION ratchet, silent-constants ratchet, role registry, model catalogs, autocloseable-closed, public-surface and constructor-width ratchets, schema-keys-consumed), Konsist laws (core OS-escape denylist, Konsist map read from the Gradle module law, stale-allowance detection) and two behaviour walls (DaemonStopOrderTest, economics contract derived from the row type). Fix rows V4-99..V4-115 burn the inventories down; the gate is red BY NAME until they land.

### Head isolation (V4-115, operator ruling 2026-09-17)
The 2026-09-16 cross-head resume linked every head's `projects/` at `~/.claude/projects`, so vanilla `claude -c` could restore a claude-deepseek transcript and warn about an unknown model. Sessions become bounded by head (real per-head trees, no shared link); cross-head resume is an explicit `-r <session-id>` copy into the calling head's tree with the model fields rewritten to that head's pinned model.

### The last three feature rows (V4-175 … V4-177)

- **V4-175 — the wizard asks which lane, and the console reads the payload the daemon sends.** `webui`'s claude-head slice had been written against a payload `/api/claude-head` was expected to serve and never did: `head`, `config_dir`, `auth_kind`, `claude_on_path`, `wrap_supported` were declared and never crossed the wire, while `resolves_to`, `real_binary_path` and `claude_logins` crossed it and nothing read them. It survived because every check took its denominator from the declaration it was checking, and `/api/claude-head` was the one route missing from the daemon's `WebuiContractTest` — now its eighth arm, so the field names are pinned from outside. Two shared bugs fell out: the console lifted `error.message` from a control plane that writes `{"error": "<sentence>"}`, so every daemon refusal reached the screen as `HTTP 409`; and a wrap action returned its own answer instead of re-reading, so a refused wrap left the page claiming a lane that was not in force. `splice setup` now asks the lane question and drives the same `POST /api/claude-head/wrap`.
- **V4-176 — the supervision contract is stated as config, not as one box's tool name.** 25 references to this machine's private supervisor and build wrapper were spread across 12 files, each written while the fact was fresh and true here. A sweep fixes the 25; the 26th arrives in the next comment. So the row restates the requirement as two knobs (`supervisorUnit`, `mcpSlice` — a box whose packager named the unit differently was reading as permanently unsupervised, which is not cosmetic: `POST /api/daemon/restart` refuses to drain an unsupervised daemon) and adds a wall that fails by `file:line`, mutation-proven on the real tree. The wall tripped on its own explanation's first draft, which quoted the name it bans — that is the check covering its own directory.
- **V4-177 — splice's state root is its own, and a pre-0.4 root is adopted in place.** `~/.claude-codex` was never chosen for splice: it is the isolated config dir of the upstream Node proxy this was ported from, kept deliberately because the paths were an external contract for an out-of-repo HUD. That consumer is gone, measured rather than argued — the live statusline names none of those files, this directory, or splice at all. The rename is one line; the work is the migration. When `~/.splice/state` does not exist and `~/.claude-codex/state` does, the old root IS the state dir, read in place: nothing copied, nothing moved, nothing to half-finish, and the decision is re-derived from the filesystem on every construction so the daemon, the CLI and `splice-launch` agree without coordinating. Both arms key off `<root>/state` and never the root, because `~/.claude-codex` is ALSO the wrapped Claude Code config dir of the head keyed `codex` — probing the root would mint the mgmt-key inside another product's config dir. `splice doctor` says which root is live and warns when a pre-0.4 one is left behind unread; `kt-state-paths-single-source` now guards both roots; and `StateDirAgreementTest` enumerates every bash copy of the rule from the tree and drives all four filesystem shapes plus the env precedence through both implementations, with a planted wrong resolver proving the harness runs the script.

- **V4-178 — an image inside a code-mode tool result is announced, not refused.** A live claudex session stalled on `500 code-mode tool result '…' contains unsupported non-text content`: agent-browser returned a screenshot as an image block inside a `splice_exec`-owned tool result, the builder refused any non-text part, and because the image stays in the client's history the refusal repeated on every following turn until compaction. Text parts are the output; an image or document part becomes an inline marker naming the media type and size and why a script reads results as text only — nothing dropped in silence, nothing dies. Carrying the pixels through the code-mode continuation is a follow-up row.

- **V4-179 — a screenshot inside a code-mode tool result reaches the model.** The follow-up to V4-178, built to splice-astra's contract and closed by their bounded review (three findings, all fixed forward). The provider renders each result's images ONCE through the dialect's own tool_result policy (`ResponsesToolResultMedia`, shared with the ordinary path), persists them on the accepted result entry (`CodeModeResultSnapshot.media`: null = a record the previous daemon wrote, `[]` = captured with no media; no metadata-version bump, v3 files load unchanged), and the canonical history replays them exactly once after the record's custom output — the same place the ordinary path puts a tool_result's images. Ownership is by exact bytes and position, so a user's own image stays the user's. The script still reads text only; its marker now says where the pixels went. A record the previous daemon wrote keeps matching after the upgrade (the turn also carries the old rendering), a second script in the same turn re-canonicalizes without doubling the image, and a completed record replayed with different media is omitted from the rewrite (nonfatal, logged) rather than mixed with the captured one.

### Repository restructure (V4-180 … V4-182, splice-vast) and head-bounded continue (V4-183)

- **V4-180 / V4-181 / V4-182 — the restructure PRs 1 and 2 (#160, #164), merged as 89c85e1e and d02bd179.** Every move is a git rename (no adds, no deletes, 2150 files before and after), the ast-grep rules live under `quality/rules/`, the build root is the repository root, and a census of the tree stops at nested git trees so a worktree under `.claude/` is never reported as an unclaimed module. Each PR carried its own gate of record (`GATE: PASS`, 200 legs) and its merge-ref CI.
- **V4-183 — a bare `-c` / `--continue` is bounded to the head.** Under the shared transcript tree (V4-64 / V4-168) Claude Code's `--continue` resumed the newest transcript in the cwd whoever wrote it and appended this head's turns into that file: nine transcripts in seven days carried two model families, seven of them deepseek sessions continued in place by a claude head. The operator's ruling: keep the shared tree, bound continue to the head. Ownership is recorded where it is decided — the SessionStart hook each head carries now fires on `startup` and `resume`, and the daemon writes `{id, cwd, transcript, at}` into a per-head index in that head's own config dir (`SessionOwnership`). The launch shim sends its cwd (shim-4), and the daemon turns `-c` into `--resume <this head's newest session in that cwd>`, or a new session said in one sentence — never the newest foreign one. `-r <id>` and the picker keep joining any session by name; a shim older than shim-4 keeps the client's `-c` and is named as the fix.

### Capability-oriented layout migration

- Move provider and dialect projects under `integrations/`, preserving directory-derived Gradle identities and module-law enforcement.
- Move documentation from `.docs/` to `docs/`. Consolidate all 19 tracked `checks/` files into their owning tools and preserve 1000 ignored runtime receipts under `tools/e2e/receipts/`. Generated ignored caches are explicitly dispositioned rather than deleted.
- Add `features/heads` with the first concrete head-start slice and feature-owned ports. Preserve authenticated routing, head lookup, lifecycle ordering, audit semantics, and response shape. This is the first extraction, not completion of the remaining capability migration.
- Update build inputs, source inventories, architecture protections, rule globs, registry homes, and relocation baselines. Coverage exclusions are checked per declared rule/module/source-set/file-glob cell with red/green regression tests.
- Fix review findings in the moved offline harness, console probe root discovery, provenance input invalidation, and test isolation. Final bounded repair review returned no real defects.

### Gate

`bun tools/gate run` under `buildgate.slice` passed on clean code commit `cf7778e7e73fbeff49d655f07757cb0b73a9abe3`, with unchanged HEAD and empty status. This is the whole ladder, including clean Gradle checks and release rehearsal, not a Gradle-only claim. The subsequent `0be5f70c` commit records the observed deployment in the ledger and changes no code.

The first remote gate exposed an inherited-selector leak in the launcher rehearsal. A container reproduction and a failing subprocess regression confirmed that ambient `XDG_CONFIG_HOME` escaped its environment scrub. The fix clears every declared selector before per-arm overrides. All launcher mutants remain effective, the release package passes 58 tests, and the repaired code commit passed [remote CI](https://github.com/torad-labs/splice/actions/runs/35631900145).

The clean-export jar passed the offline fresh-machine harness, 31/31 steps. It was installed after backup, and the restarted daemon's open jar descriptor matched the built artifact SHA-256:

```text
7a678ce7621a393a27a9a44f0ce92791dd9d3568a826800ff987604d34bbff56
```

Live health reported all 10 configured heads ready and no failed heads. The runtime still warns that installed Claude Code 2.1.278 is newer than the tested 2.1.257 pin. Deployment evidence is recorded in `.dev/campaigns/repository-layout.toml`; the broader migration remains `in_flight`.

### The deferred row, landed (V4-141)

- **V4-141 — TCP keepalive on upstream sockets, via the OkHttp engine.** `java.net.http.HttpClient` exposes no socket options (JDK-8338681), so for as long as it was the upstream engine a half-open upstream connection read as healthy for the whole turn cap — measured, not argued: the daemon's established upstream sockets showed no keepalive timer. The engine is now ktor's OkHttp, for the one reason that it takes a `SocketFactory`: every upstream socket is armed before it connects (SO_KEEPALIVE, TCP_NODELAY, idle/interval/count = 30s/10s/3, a dead peer known within 60s), which also closes G26. The swap's two costs are paid and pinned rather than met in production. OkHttp's Dispatcher caps five requests per host by default, which would serialise every turn against one provider; it runs at 4096 on a virtual-thread executor. And ktor's OkHttp engine reads each response body with a blocking `source.read` on its engine dispatcher, `Dispatchers.IO` by default — the first run of the swap deadlocked the 1000-stream load test at exactly the ramp width (96 held at the mock, 0 live at the client) because 96 held streams pinned all 64 IO threads and the head could not forward the deltas it already had. The engine dispatcher is the same virtual-thread executor, both are pinned by `UpstreamTransportOkHttpTest`, and the load test is green again at the same concurrency (`n=1000 all_live=true peak_concurrent=1000 clean_finishes=1000 errored=0`). Six components entered in `verification-metadata.xml` by hand, every sha256 cross-checked against Maven Central.


marcospaulo:
### Repository layout (LAYOUT-01): capability slices, installed at 99e8937e

- **2d740991** `refactor(layout)`: completes the `client/ → integrations/claude-code` wiring that c0af51b0 carried without its settings (the `:client` configuration failure in run 35738760190). Lands the capability boundaries:
  - `features/turns`
  - `features/sessions`: the Claude Code `TranscriptReader` implements the sessions-owned `SessionTranscripts` port.
  - `features/usage`
  - `features/heads`
  - `features/models`
  - `integrations/http` and `integrations/mcp`
  - `LogSafe`, the terminal palette and `ActivityDays` moved to `core`.
- **095cab0b** `fix(gate)`: the no-python wall no longer charges the restructure census for naming `python-burndown.json`. Measured: exactly one charge dropped (41 → 40), with a red-green pair.
- **99e8937e** `fix(test)`: the code-mode worker-cap test waits for the closed worker's permit, fixing a race that went red under full-ladder load.
- **Verification:**
  - Every relocation is recorded in `.dev/restructure/capabilities.tsv`.
  - Rule selection is preserved: of 9,576 original rule×file pairs, 0 were lost. The 4 bridge pairs that moved now fall under the stricter `runBlocking` ban.
  - `prove-coverage` passes 63/63.
  - Whole-build `check` is green.
  - `bun tools/gate run` reports **GATE: PASS** in a clean worktree at 99e8937e.
  - CI is green on the same sha.
- **Installed:** jar sha256 `e3f94eb4…`, matched against the open file descriptor after the restart.

marcospaulo:
**Hook audit (febea02d) and the red-CI fix (caae661b)**

`febea02d` removes the repo hooks that cost more than they caught. Each was measured by per-event latency and by its blocks over 14 days of transcripts:

- The SessionStart law dump: 152 KB, about 38k tokens, into every seat at every start and compaction. Laws still print in each row's `packet`.
- The Stop and SubagentStop whole-tree ast-grep scan: 2.1–3.7 s per turn end. Its 12 blocks were all findings in files the blocked seat had not touched.
- `gradle-stop-guard`: a bare `./gradlew compileKotlin` outside the build slot on every Stop.
- The in-flight warning repeated on every Stop.
- The one-committer guard: 406 marked git writes and 0 blocks in 14 days.
- Two dispatchers with no modules.

The write-time wall, no-python and the ledger raw-edit guard stay. The in-flight re-anchor stays, now only for a seat that claimed a row.

`caae661b` fixes CI, red since c6ee5eae: V4-87 documents `TERM`/`NO_COLOR`/`COLORTERM`, and V4-88 lowers the silent-const census to record the StatusTable win.

CI (`ci` including the complete repository gate, plus `coverage`) is green on ca34a0b2, which contains both commits.

marcospaulo:
## Console: every page against the real daemon, and a Playwright leg in the gate (web-console M3R-01)

Three commits on this branch: f5c04377, bb9a81c8 and 6b328525.

**What was broken.** The console's types were written against payloads nobody had read, and its unit tests were fed fixtures shaped like those types. So the two agreed with each other and neither matched the wire. Against a live daemon on 2026-09-22:

- accounts crashed to the router error screen with `e.windows is not iterable`;
- turns sent a 400 on every poll because it asked `/api/perf/turns` without a `head`;
- doctor printed `undefined` for rollback;
- models and fleet read head fields `/api/models` never sends;
- logs polled a hard-coded head.

**The fix (f5c04377).** Each read is typed as its real wire shape, with adapters at the entity boundary: `entities/account/model/wire.ts` and `entities/perf/model/turns-wire.ts`. Turns fans out one request per head. `/api/perf/turns` keeps its head-required contract, which `ConsoleRoutesTest` pins, and a head that fails is named rather than dropped silently. 6b328525 keys the models detail by `{head, id}`, so two heads serving one model id each show their own topology windows.

**The suite (bb9a81c8).** `console/e2e` boots the fat jar in isolation with:

- a mock upstream that sends SSE and `x-codex` quota headers;
- a pooled OAuth head, a solo head and an api-key head;
- one real turn driven through them.

It then opens every page the router globs. A page fails on:

- the router error screen or the lock screen;
- an empty `main`;
- `undefined`/`NaN`/`[object Object]` in the text;
- an uncaught error or a console error;
- an `/api` read the daemon refused.

Four journeys check the values the stack put there: the turn, the 42%/7%/plus account windows, the pinned models, and the topology's 300k head window.

- Red first: against the pre-fix bundle, 6 tests failed on exactly those defects.
- Now: 19/19 against the jar's own bundle, locally and in CI: gate run 35814062636 at ca34a0b2 reported GATE: PASS with consoleE2e 19 passed, consoleWireKeys PASS and selftest 59 arms.

**Gate.** Four new ladder legs, all behind `:app:shadowJar` except the replay test:

- `consoleE2e`;
- `consoleWireKeys` (the V4-140 probe, which existed but ran in no leg);
- `consoleWireKeysSelftest` (59 mutation arms over 25 routes);
- `consoleWireKeysReplay`.

`:console:lint` and `:console:typecheck` now cover `e2e/`. CI installs the lockfile's chromium headless shell before `npm run gate`.

**Filed, not fixed here** (ledger followups in `.dev/campaigns/web-console.toml`):

- **Unbuilt surfaces blame finished daemon rows.** Some console surfaces are still unbuilt while their empties name daemon rows that are done: the fleet detail's restart and pool panels, and the team board's economics and activity panels.
- **The coverage manifest is stale.** 24 of its 26 `pending` route dispositions name routes the daemon registers.

https://claude.ai/code/session_018o5oN9dmpUcDLkyZf8YhLZ

marcospaulo:
## Console: every served route the pages were missing (web-console M4-01..08)

Found by diffing the daemon's 58 `/api` registrations against every console `request()` call. All eight rows are done on the combined tree `72597fd66`.

| row | what the page does now | commits |
|---|---|---|
| M4-01 | Teams reads chat, activity and economics and writes create, replace and session bindings, each typed as its wire. The board's views are built from the live payloads, not fixtures | `6893780c1` `d8be88292` `e9247a255` |
| M4-02 | The fleet head detail shows its account pool and has one draining-restart control (refused when unsupervised) | `735873074` `7865b5279` |
| M4-03 | The doctor playground sends a prompt through a head; the upgrade section restarts the daemon | `a563e88f0` `5f9572f51` |
| M4-04 | A per-head body-capture switch (PUT, then re-read) | `0671c96e6` |
| M4-05 | The project detail, fleet-wide session edges and compaction rules are read as their wires | `bb3db72ea` `0cb068fd0` |
| M4-06 | The coverage wall parses what the daemon serves from the main Kotlin sources and fails by name on a served route left pending or undispositioned; the stale "not built / V4-xxx" empties are gone | `aee915e43` `af3ebed52` |
| M4-07 | Nothing from a doctor report the redaction check refused reaches the screen; the daemon's own `<redacted>` mask is not treated as a leak | `5d5cfd71c` |
| M4-08 | Accounts marks each pool's next target from the daemon's own `next_target`, in the daemon's order (pin, primary, sticky, lowest 7-day); the 7-day slot is matched the way Quota.kt files it | `f3495ccdc` `2721a8c1d` |

**Verification on `72597fd66`:**
- vitest: 48 files, 858 passed.
- Wire probe: PASS against a booted daemon with a seeded team; its selftest also passes.
- Playwright: 28/28 against the built bundle on a private copy of the jar. `f9e468943` fixed a harness race where a shadowJar build in the shared tree rewrote the jar mid-run and produced 500s.
- Each new rule was mutation-checked and goes red: the own-session tally, the `to_slot` join, the transcript-missing path, the 604800 s slot and the served-but-pending wall.

**Follow-ups on `fix/sse-drain`, awaiting its gate of record:**
- `bf03e3c7d` fixes the load-only frame loss that failed `HeadServerCompactCapTest` and `HeadServerFoldTest` NF-03. When the head's body threw, Ktor cancelled the response channel and dropped flushed bytes the engine had not read yet. The fix is `SseResponse`, which closes the channel cleanly in a finally, plus the `kt-head-sse-drains-on-exit` wall.
- `2df8d89ce` exempts the doctor-gate test's planted canary from org-secret-scan's supplementary pass, the red on `be05c8ab8`.
- `bed19c820` drops a type union M4-08 made redundant.

**Still open, none of them console work:**
- `GET /api/projects/{id}` does not carry the compaction scope or statusline roots yet (daemon side).
- The waterfall strings `drawer` and `showCapture` are unused. That dead code was already there; I left it.
- WD-06/07/08 still read `blocked` in the ledger although V4-132/129/133 are done. The evidence is on the M4-06 row; flipping them is the orchestrator's call.

marcospaulo:
**Console gaps closed and installed** (fix/console-gaps @ `cc62e87d0`, landing on this branch through `679f1954f` once its gate passes)

- **M4-09, the project governance fields:** `/api/projects/{id}` and each list row now carry:
  - `compaction`: the rules a compaction in that repo resolves to. `CompactionInstructions.rulesFor` is pinned both ways against `resolve()`.
  - `statusline_roots`: per head, the trusted root its statusline probes the repo under.

  The project detail prints both, through a compaction-rule widget shared with the compaction page.
- **M4-10, perf history kept:** V4-133's archive was built but never wired, so every 64 MB rotate still discarded the generation before `.1`. Heads now archive into `<state>/perf-archive`, the perf readers read it oldest first (skipping, unopened, any generation older than the window), and team economics tally from the team's creation. Wall `kt-perf-history-archived` keeps it wired.
- **WD-06/07/08:** done by supersession under V4-132/129/133. Each receipt reruns that surface's route checks and tests.
- **Also in this tip:**
  - the SSE frame-loss fix `bf03e3c7d` (HeadServerCompactCapTest / NF-03);
  - `2df8d89ce`, which clears the org-secret-scan supplementary red here;
  - management routes that throw now answer a JSON 500 naming the failure's kind and log it, where before the daemon had no SLF4J provider and they left no trace.
- **Verification:**
  - gate of record PASS on `53f9f06c0`, console e2e 28/28;
  - installed: jar `4e065af8…`, pid 3312447, backup `splice.jar.bak-20260923-pre-53f9f06c0`;
  - live checks: 10 of 10 heads ready, and the claudex picker still reads the backend.

